PR #6, opened 2016-12-17
Fix build

tkurki:
See commits for details.

This fixes the build so that `npm run build` produces a bundle.js, but I did not even attempt to use it in any way.

PR includes only the package.json changes, not an updated bundle.js.


tkurki:
ping @rob42

PR #22, opened 2017-07-02
fix: use mmsi for ownVessel if available

sbender9:


rob42:
This may require a similar change in the signalk-java-server. Otherwise the self vessel will not be found...

sbender9:
Would to help to check for uuid first?

rob42:
yes I think so, so the resolution order first>last is uuid>mmsi>url

PR #46, opened 2018-09-12
Port to Angular framework

panaaj:
Port of freeboard-sk to the Angular framework.


tkurki:
Please update the PR description with a description of functional changes and anything else noteworthy included in this PR.

package.json
```diff
@@ -1,48 +1,54 @@
 {
-  "name": "@signalk/freeboard-sk",
-  "version": "0.0.4",
-  "description": "Openlayers chartplotter implementation for Signal K",
```
tkurki: Why no description? This is shown in server App store.

package.json
```diff
@@ -1,48 +1,54 @@
...
+    "e2e": "ng e2e"
   },
-  "author": "robert@42.co.nz",
-  "license": "Apache v2",
-  "keywords": [
-    "signalk-webapp"
-  ]
+  "private": true,
+  "dependencies": {
+    "@angular/animations": "^6.1.7",
```
tkurki: Aren't all these devDependencies? Like if you install from npm (like app store does) no deps should be needed, as this is just a pre built web app, right?

tkurki:
There are now two ways to pick the chart layers shown, in the left menu and via the layers button. Do we need both?

tkurki:
Selected chart layers should be persisted in localstorage.

tkurki:
n2k-from-file does not show own ship nor self data.

panaaj:
### Included in this PR:

- Application rebuilt using the Angular Framework v6

#### Charts:

In addition to the display of online maps i.e. OpenStreetMap, OpenSeaMap, and WORLD chart outline, MBTile maps hosted on Signal K server using the *@signalk/charts-plugin*

#### Resources:

    - List and view Route and Waypoint resources hosted on Signal K server
    - Select and set a Route as the active route.
    - Edit Route / Waypoint properties *(Name, Description)*
    - Draw and save a Route
    - Add Waypoint at the Cursor or Vessel position 
    - Delete Waypoint(s) / Route(s) 

For resources to be added / updated on the Signal K server requires a resource provider plugin to be installed *e.g. GPXLoad*

#### Map Display:

    - North-up or Vessel-up 
    - Moving Map or Moving Vessel
    - Vessel Heading / Bearing lines
    - Wind true direction / apparent angle lines
    - Measure distances

#### Alarms:

    - Display both Anchor drag and depth notifications received from the Signal K server using visual and audio alarm indication as specified by the received *Notification* message 
    - Integrate with *signalk-anchor alarm* plugin to enable setting anchor drag alarm radius and to raise / drop anchor.

#### Integration: 

Allows the selection of applications installed the Signal K server to be used for:

- **Instrument panel**: *(default: `@SignalK/InstrumentPanel`)* to be displayed in the instrument panel drawer.
- **Resource Manager**: *(default: `none`)* to provide a link to an application to be used for resource management.

package.json
panaaj: NPM package is the group of files in `dist/freeboard` in which teh package.json is:
```
{
    "name": "@signalk/freeboard-sk",
    "version": "0.1.0",
    "description": "Openlayers chartplotter implementation for Signal K",
    "keywords": [
        "signalk-webapp"
    ],
    "repository": "https://github.com/SignalK/freeboard-sk",
    "scripts": {
        "format": "prettier-standard 'src/**/*.js'",
        "test": "mocha"
    },
    "author": "robert@42.co.nz",
    "contributors": [
        { "name": "panaaj@hotmail.com" }
    ],
    "license": "Apache-v2",
    "dependencies": {
    },
    "devDependencies": {
    "mocha": "^5.0.5",
    "prettier-standard": "^8.0.0"
    }
}
```

package.json
panaaj: This is the angular project package.json.

NPM publish should be done from `dist/freeboard`

panaaj:
> There are now two ways to pick the chart layers shown, in the left menu and via the layers button. Do we need both?

There is no layers button on the map, only the one in the left menu...

Not sure how you are viewing the webapp but you just need to serve  the `dist/freeboard` folder

panaaj:
> Selected chart layers should be persisted in localstorage.

selections of resources, charts and settings are all persisted in localstorage. The only non-persisted item at this point is the vessel trail as I am waiting to see how the historical data display will be handled first.

image

panaaj:
> n2k-from-file does not show own ship nor self data.

All received vessels.uuid messages are processed to form either the "self' vessel display or the 'additional' targets on the map.

package.json
```diff
@@ -1,48 +1,54 @@
...
-    "mocha": "^2.3.4",
-    "nanomodal": "^5.1.0",
-    "openlayers": "^3.16.0",
-    "simplify-js": "^1.2.1",
-    "uglify-js": "^2.4.20",
-    "watchify": "^3.1.2",
-    "web-audio-daw": "^2.3.1",
-    "xml2js": "^0.4.16"
-  },
+  "name": "freeboard",
```
tkurki: @signalk

package.json
```diff
@@ -1,48 +1,54 @@
...
+    "start": "ng serve",
+    "build": "ng build",
+    "test": "ng test",
+    "lint": "ng lint",
+    "e2e": "ng e2e"
   },
-  "author": "robert@42.co.nz",
-  "license": "Apache v2",
-  "keywords": [
-    "signalk-webapp"
```
tkurki: Keyword is a must here.

tkurki:
Sorry, my bad, I was looking at the old version, I did not even realise how much work you have done! Silly me...

panaaj:
All good... it all should all work on the server as it is expecting to be run from the server..  I tested the files in the PR again on my Docker and they worked fine.... you made me nervous that I'd uploaded an incomplete set....

README.md
```diff
@@ -1,45 +1,82 @@
...
+
+This is an Angular project generated with [Angular CLI](https://github.com/angular/angular-cli).
+
+1. Clone this repository
+
+1. Run: `npm install` *(to install Angular CLI and project dependencies)*
+
+1. Run `ng serve` to start a develeopment web server and navigate to `http://localhost:4200/`. The app will automatically reload if you change any of the source files.
+
+* **Note:** The served application will look to connect to Signal K server at the ip address defined in the `DEVHOST` property in the `src/app.info.ts` file.
```
tkurki: DEVHOST is apparently used even when used running off dist produced with `ng build`.

panaaj: Yes that is correct... you need to ng build —prod for a production build.
Ng build allows you to test in dev mode on a local web server without compiling in the web browser

rob42:
@panaaj First let me say this is great work!. Its been needed for a long time and Ive just not had the time. You have made excellent progress, and I'm looking forward to seeing this replace the current version.

I dont think its ready to merge into master yet as there are a few build/deploy/etc issues to sort out, so Ive created an panaaj-angular branch in here and merged it into that. This will allow us to bring the app to the point where it can replace the current freeboard without disturbing the current users (who update from master)

So future pull requests can go to panaaj-angular. I will give you commit rights to the branch so that you can work directly.

rob42:
Another note. When I run this under artemis-server it tries to load freeboard-sk/index.html, which fails as its the old version. To run the app I needed to execute freeboard-sk/dist/freeboard/index.html. As @tkurki  mentioned above the node server/npm default is /dist, but not all apps follow that. So if we include an index.html file in /freebaord-sk that sets basedir and redirects everything should always work.

rob42:
Next item, when run against the artemis-server (with security enabled) your app tries http://localhost:8080/loginStatus, which is 404. I guess its on the node-server. @tkurki is this signalk spec or node-server specific? Either way we need a std signalk way to do this.

The next problem is that with no jwt token the artemis server redirects the browser to http://localhost:8080/login.html, which also fails. Again I think we need a std here so all apps can authenticate ok.

panaaj:
> Another note. When I run this under artemis-server it tries to load freeboard-sk/index.html, which fails as its the old version. To run the app I needed to execute freeboard-sk/dist/freeboard/index.html. As @tkurki mentioned above the node server/npm default is /dist, but not all apps follow that. So if we include an index.html file in /freebaord-sk that sets basedir and redirects everything should always work.

The `dist/freeboard` folder is the output destination of the angular project which contains the `deployable application files`. These are the only files that should make it to the server and is where `npm publish` should be run from.

This location can be easily changed to an appropriate folder name... The contents of this folder (which includes index.html) can be placed in whatever folder structure is suitable. 

For example, I work and test with the node server by copying the contents of the `dist/freeboard` folder to a folder on the node-server  (e.g.`/home/signalk/freeboard`) and `npm link` it from `/home/signalk/.signalk/` to make the application appear in the `webapps` list of the admin ui where the application can be launched and used successfully.

Let me know what the folder should be named and I will adjust the project config file to make it so.

panaaj:
> Next item, when run against the artemis-server (with security enabled) your app tries http://localhost:8080/loginStatus, which is 404. I guess its on the node-server. @tkurki is this signalk spec or node-server specific? Either way we need a std signalk way to do this.
> 
> The next problem is that with no jwt token the artemis server redirects the browser to http://localhost:8080/login.html, which also fails as it doesnt seem to open the page in the browser. Again I think we need a std here so all apps can authenticate ok to any server

This is one of a number of the questions I had during development regarding how to approach certain situations... I discovered the `/loginStatus` url from trolling the source code of the server not though any documented guidance so at this point I only make provision to detect what the server configuration is.
I figured that you need to be logged in to launch the application from the server console anyway.

Happy to align to whatever the proposed process / sequence should be.

sbender9:
Definitely should not be using /loginStatus. 

I’ll take a look at the code tomorrow. 

Definitely need to get the security stuff into the spec. Just have not gotten around to it yet...

panaaj:
It seems to me that there may be significant enough differences between node and java server versions, so how can I ensure that I create an application that is suitable for both without having to create two applications?

I have taken the approach that the app is a window to the information provided by the server, this will ensure multi-user scenarios display consistent data (e.g. in the wheelhouse and on the bridge) and all data is relayed back to the server.

To do this the app relies on the server to provide the necessary functionality either natively or via plugins to serve / accept the required data:
-  Alarms, *(using signalk-anchor-alarm for anchor watch)* 
- Resources *(using GPXLoad for routes / waypoint management)*
- Charts *(using @signalk-charts plugin)*. 

So a consistent server-side api would be best, so in the case of **anchor watch**, the plugin provides `raiseAnchor`, `dropAnchor`, `setAnchorPosition` apis to drive it and then relies on the plugin to generate the notifications which the app will then display.  

As an application developer can I assume that the same plugins / apis can be made available on the java server?
Is this approach going to be an issue? 
Am I barking up the wrong tree?

I'm trying to look past the `single-user` use case and the `what works for me on my boat` to broader use cases. **It is a server after all!**

sbender9:
As an app developer you really need to look at the specification. If something is not in the spec then you can't assume that all servers will have it.

I don't think that means that you should not use functionality that's available but not in the spec, you just need to be smart about it. If you call `dropAnchor` and get an error response, then let the user know that the server does not have an anchor alarm.

In general, node server is going to be far ahead of the spec. We often try things out there before we propose changes to the spec. And changing the spec can take time. In WilhelmSK, I try to make sure that I can support a server that purely follows the spec, but also support more cutting edge features gracefully.

Please join us at https://github.com/SignalK/specification and raise issues or make pull requests. 

In this case, I'd say it's up to @rob42 if there should be code here that uses non spec compliant features.

panaaj:
> Definitely should not be using /loginStatus.
> 
> I’ll take a look at the code tomorrow.
> 
> Definitely need to get the security stuff into the spec. Just have not gotten around to it yet...

In the spirit of disclosure here are other urls used to detect the state / capabilities offered by the server and provide a better user experience...hopefully these are ok to use:

- `/webapps` used to get a list of installed apps to present so the user can select their preferred instruments or resource manager.
- `/plugins` used to detect the presence of required plugins to provide contextual feedback to the user and *trim* the UI of functions that are not available.

panaaj:
Errors are handled gracefully.

I know documentation is not something that everyone runs to do, but it would be good (and might prevent a lot of wasted time answering the similar questions over and over) if there were a simple `Signal K application do's and don'ts` list to outline the key things that will ensure an application aligns to the intended operation of the Signal K eco-system.

Speaking as someone who has come in only recently, I have spent a lot of time trying to learn the "Signal K philosophy" by scanning code and scouring the spec to ensure what I produce conforms to good practise .  I know that the answers are not always going to be there but the lack of some basic guidance then leaves me, the developer, to my own devices.... not sure you want that!!

tkurki:
public is good for node server. Afaik Artemis mounts the root so index.html there that redirects to ./public/ will do the trick.

To simulate how the app will be installed from npm you can run `npm pack`. This will produce .tar.gz that you can check manually or cd to an empty directory and npm install <thetarfile> to check what is actually installed. That should not install any dev time deps, just the Angular app files. Adjust .npmignore as needed.

rob42:
The issue is not really the dir name but the lack of a predictable layout for apps, and predictable functionality provided by the server. Thats not a fault of your app, its a signalk thing. 

One of the reason for maintaining at least 2 signalk server implementations is that it highlights these issues, so that we build a std that is truely interoperable, not just a js/node framework. While the UI is almost exclusively browser/node-server/java server now, its important that a C++ app on an MFD can also be fully functional, or any future development. 

The artemis-server treats all apps as static files, and for simplicity unpacks them into /signalk-static/APP_NAME/. It therefore assumes there is an index.html file there as a bootstrap. The reason wasnt anything more than being the simplest solution, and the most generic in terms of whatever web-server is used. 

If we go the npm way (which has its own merits) then thats great for node-js, but introduces complexity for other web servers (nginx,apache,tomcat, etc) as they dont use those formats by default, and have yet more issues with the shared node-modules/ paths etc. 

So we should decide on a generic app format, that can be installed on any web server successfully. Likewise we need predictable urls for login etc.

As a starter for discussion we should specify that any signalk app can be retrieved and unpacked into an APP_NAME  directory on any web server (apache, nginx,tomcat, whatever), and should run when /APP_NAME/ is invoked. That assumes that all dependencies are in the bundle or auto-resolved. 

Likewise it seems that an auth requirement should redirect to /login, and /logout. Artemis-server is using JWT's, we might need to specify how an app applies security internally, artemis applies it against the signalk message so all data is always protected no matter the source.

rob42:
> It seems to me that there may be significant enough differences between node and java server versions, so how can I ensure that I create an application that is suitable for both without having to create two applications?

_The reason for at least two is to force compatibility issues to the surface :-)  Yo need to assume that only documented signalk messages are available, and that documented urls are available. We encourage people to suggest additions to suit their needs, thats how we grow. The web related urls certainly need attention as this has demonstrated._

> 
> I have taken the approach that the app is a window to the information provided by the server, this will ensure multi-user scenarios display consistent data (e.g. in the wheelhouse and on the bridge) and all data is relayed back to the server.
_Yes, this is how I see it. We need to add the urls that provide required web functionality now, eg /webapps, /login, etc _
> 
> To do this the app relies on the server to provide the necessary functionality either natively or via plugins to serve / accept the required data:
> 
>     * Alarms, _(using signalk-anchor-alarm for anchor watch)_
> 
>     * Resources _(using GPXLoad for routes / waypoint management)_
> 
>     * Charts _(using @signalk-charts plugin)_.
> 
_if these plugins are running at the browser to create signalk messages for the server, then thats fine, but we cant assume that you can send a GPX to a server plugin and get back the converted response as the plugins you reference are nodejs specific. _
> 
> So a consistent server-side api would be best, so in the case of **anchor watch**, the plugin provides `raiseAnchor`, `dropAnchor`, `setAnchorPosition` apis to drive it and then relies on the plugin to generate the notifications which the app will then display.
> 
_Here my view is different. We dont need urls for the misc vessel tasks. To set the anchor position send a signalk json message setting the appropriate keys using a PUT or POST. The server will then echo the change to all subscribed clients (the multi-user bit). There is no need for a specific url for this and if we did,  full functionality would then require a set* get* url for every signalk key.

Think of the server as a state cache. To raise the anchor (we dont have a key yet!) set `anchorCapstanState=up`, and the server will echo it to the capstan controller (which is subscribed to the key) and the winch will start or whatever. When the fuse pops, it sets `anchorCapstanState=fail` or whatever and thats echoed to your app by the server. If all action is performed by signalk messages then its also possible for the bow remote to participate even when it just has a simple serial connection and doenst know http._

> As an application developer can I assume that the same plugins / apis can be made available on the java server?
> Should there multiple server types?
> If so how can we ensure alignment?
> Is this approach going to be an issue?
> Am I barking up the wrong tree?
> 
> I'm trying to look past the `single-user` use case and the `what works for me on my boat` to broader use cases. **It is a server after all!**
_Definitely aim to support the multi-user cases. See above,  works well, the pattern is basically asynch messaging, using the signalk server as message server. So when we want an action we dont try to do it directly, we simply send a message setting the required state, then we wait for the system to adjust to our requirement (or it can set an error/notification).  _

panaaj:
So.... taking all of the above in to account here is what I am thinking....

1. From an **application installation** point of view it seems the presence of the Angular project source at the root level is a level of complication that can be avoided so I propose I change the structure to the following:
**/source** - angular project source
**/** - runnable app produced by angular project build `ng build --prod`
.npmignore - excludes /source folder

I think his should suit everyone.

2. From an application operation standpoint it seems that the following need to be clarified / addressed:

**Resources:**
The acquisition of resources from the server should be no issue as they are retrieved from `resources/routes` and `resources/waypoints`. 

As for writing resource data back to the server: the app can be configured to use `HTTP PUT requests` or `delta updates` to create, edit and delete resources. Therefore as long as the server responds to these resources should be managed correctly.

For the node-server this will require a plugin like `GPXLoad` which provides that endpoint so resources are persisted. Not sure what functionality of the Artemis server is in this space.

**Charts:**
The acquisition of charts from the server should be no issue as they are retrieved from `resources/charts`.
For the node-server a plugin is required to service this path. Not sure what functionality of the Artemis server is in this space.

**Anchor Watch:**
This is where we have departure from the views above... I took the approach of using an available plugin because that was the only way I could get the functionality to work.
This is clearly not the best approach so for this release we can either:
- Remove the Anchor Watch function for now
- Agree how it should interact with the server and I adjust the app accordingly
- Leave it for now, the UI looks for the presence of the plugin using `/plugins` and does not display the option, and fix it later.

Thoughts?

rob42:
We should add the following std urls as web api, bit rough yet but a starting point. They are under `/signalk/function` so we can filter their access

We already have:
```
/signalk/v1/stream
/signalk/v1/api
```
Add:
```
GET /signalk/v1/apps/list - use current npm/node-server json format
GET /signalk/v1/apps/install/{id} - ditto
GET /signalk/v1/apps/upgrade/{id} - ditto
GET /signalk/v1/apps/remove/{id} - ditto

POST /signalk/authenticate/login - post json {"id": "", "password": "" } - hashed?
GET /signalk/authenticate/logout/{id} 

GET /signalk/v1/logger/list?logDir={}
GET /signalk/v1/logger/get?logDir={}&logFile={}

POST /signalk/v1/upload/chart/{format} - upload chart resources, eg tiles, multipart
```

sbender9:
@rob42 I think we're missing something in the spec here RE:. anchor watch. Only being able to PUT the anchor position and radius is not enough and limits what the server can do. 

For example, my anchor alarm plugin allows you to enter the rode length and then set the alarm. This then uses your depth, position and heading to estimate the anchor position. And it actually works really well.

This does not really fit into just being able to PUT/POST individual path/values.

rob42:
Its still ok. (We need to add `anchor.rodelength` if we dont have it) Your app just sets the rodelength, and turns anchorwatch on, and the server does the calc on actual anchor position.

rob42:
The java server did a different position optimization. As you drifted around your anchor it trimmed away all the area the anchor could NOT be (since its one rode length from the boat). so it worked out where the anchor actually was over time.

sbender9:
How do you "turn anchorwatch on"?

rob42:
Ive added rodeLength below:
```
"anchor": {
      "type": "object",
      "title": "anchor",
      "description": "The anchor data, for anchor watch etc",
      "properties": {
        "source": {
          "description": "Source of this data",
          "$ref": "../definitions.json#/definitions/source"
        },
        "timestamp": {
          "description": "Timestamp of the last update to this data",
          "$ref": "../definitions.json#/definitions/timestamp"
        },
        "maxRadius": {
          "$ref": "../definitions.json#/definitions/numberValue",
          "description": "Radius of anchor alarm boundary. The distance from anchor to the center of the boat",
          "units": "m"
        },
        "currentRadius": {
          "$ref": "../definitions.json#/definitions/numberValue",
          "description": "Current distance to anchor",
          "units": "m"
        },
        "position": {
          "description": "The actual anchor position of the vessel in 3 dimensions, probably an estimate at best",
          "$ref": "../definitions.json#/definitions/position"
        },
        "rodeLength": {
          "$ref": "../definitions.json#/definitions/numberValue",
          "description": "Rode length of anchor.",
          "units": "m"
        },
      }
    },
```
setting the alarm is done by changing meta on the `currentRadius`
```
{
                           path: 'navigation.anchor.currentRadius.meta',
                           value: {
                               "displayName": "Anchor Watch",
                               "shortName": "Anchor Watch",
                               "warnMethod": "visual",
                               "warnMessage": "Check  anchor, near watch limit",
                               "alarmMethod": "sound",
                               "alarmMessage": "Dragging anchor",
                               "zones": [
                                   [0, 45, "normal"],
                                   [45, 50,  "WARN"],
                                   [50, 999999, "alarm"]
                               ]
                           }
                       }
```
The server should already have a process that monitors all meta zones, and that triggers the appropriate alarm. To turn it off again set
 "zones": [
                                   [0, 999999, "normal"]
                               ]

sbender9:
But now it's up to the client choose the alarm radius (zones), right?

rob42:
Yes, but setting anchor watch on and deciding on radius go together. Having a default is a handy thing though. Freeboard-sk had a slider set at last value, and you changed as needed.
Another option for that is to change the meta state, not radius. eg 
```
                        "zones": [
                                   [0, 45, "normal"],
                                   [45, 50,  "normal"],
                                   [50, 999999, "normal"]
                               ]
```

sbender9:
So my example above is still not possible. (also, still no way for a client to change meta data)

The server should have the smarts here, not the client. I think you're making it way too hard for a client to set an intelligent anchor alarm.

sbender9:
We should move this somewhere else, I'll raise an issue in specification tomorrow...

tkurki:
GMTA. Adrian, you got a bit more than you bargained for.

I started with https://github.com/SignalK/specification/issues/504

panaaj:
Oh well.... nothing is ever quite as straight forward as one hopes! 

I thought that the "leeway" I could give myself as a developer based on the number of unanswered questions I had would lead to this. 

It's different when you are building something that is closer to the core rather than just an app or plugin I dreamed up that might be useful.

Maybe it says something about me but I enjoy the discussion... that's when good things happen!

tkurki:
Yep, some of this has been long overdue and I am really glad that we are discussing these things. Let's turn this stuff into docs & implementations!

panaaj:
We are making progress!! 
Updates to node-server anchor alarm have pretty much aligned server operations so a notification can be displayed.
There might be some further minor tweaks required but it is looking good.

To set alarm:
1. HTTP PUT navigation.anchor.position
2. HTTP PUT navigation.anchor.maxRadius

To turn off:
1. HTTP PUT navigation.anchor.position= null

image

panaaj:
Added functionality to `drag and drop` or load via menu option resources from GPX files.
This should complete the alignment of functionality from the previous version.
image

panaaj:
Have pushed some updates to the `panaaj-angular` branch that provide the following behaviour:

1. Detects server type and uses the result to:
- Display relevant welcome message - i.e. for node server advising that required API paths may need to be serviced by plugins
- Use appropriate url for retrieving list of installed apps to present for use in the Instrument panel i.e. for node `/webapps`

2. Removes use of non-API paths i.e. `/dropAnchor`, `loginStatus` for application operation.

3. Persists vessel trail to IndexedDB and fixes some minor display issues and is now very much feature complete.

It's now probably in a state where it would benefit from some wider testing to unearth further issues.
Not sure if a Slack post in `apps_and_clients` section requesting volunteers to test this branch is appropriate?
Given Freeboard is normally included in the node server installation I'm guessing there needs to be a reasonable level of testing.

rob42:
I think its time to merge this into the master branch. I will make a legacy branch for the old code.  Then I guess if we publish it to npm it will go into new installs by default, and existing node users can upgrade.

Might need @tkurki help to make both versions available in npm somehow.

tkurki:
Old versions are not overwritten by a new one, just that you will get a new one by default.

We do have the option of pinning the version that gets installed when you install the server - then the new version will show up in app store updates and the user can choose to install it.

tkurki:
@rob42 I am not too crazy about merging a PR that has unresolved review conversations. Very unmotivating.

Now that you rushed this to master are you going to take care of seeing this through?

I am unclear on how exactly this should be built and published, I wanted to work the details out in the branch before merging. Details related to that
- I prefer `npm version` in the root folder for managing versions
- none of the scripts produce a production build
- if publishing is not `npm publish` I'd prefer `npm run publish` or something along those lines

tkurki:
- `npx ng build --prod`
- `npm pack` in dist/freeboard
- `npm install ..../signalk-freeboard-sk-0.1.0.tgz` in node server dir
After this the new version loads, but it ends up in the browser window freezing quickly.

rob42:
Reverted again.

rob42:
Hmm..cant reopen this pull request and dont want to loose the history above, any-one know how to resurrect this one?

rob42:
@tkurki sorry about the unresolved reviews, they were way up top and I didnt notice them. They seem to be old now, can you resolve if you are happy with subsequent developments?

package.json
panaaj: Folder restructured to move angular project source to the `source` folder and the deployable files are located in the `root` .npmignore added to exclude `/source`

package.json
panaaj: Folder restructured to move angular project source to the `source` folder and the deployable files are located in the `root` .npmignore added to exclude `/source`
Application `package.json` file is now in the root folder.

rob42:
> @tkurki I added comments and resolved the comments... hope that's OK.
> 
> Also @rob42 I noticed a change to the `tilemapUrl` for chart display... there may me a misalignment here with the @signalk-charts-plugin.... I would expect that the value from the API path is the complete path to the resource and does not require any adjustment in code.
The url is the root of the tilemap, so you still need to append the x.y.z part or you get a zillion 404's on just the root dir. 
The tricky bit is that there are subtle differences between web tile map formats. See https://en.wikipedia.org/wiki/Tiled_web_map The formats I use (TMS?) needs z.x.-y.

I suspect we need to support different ones, which means adding `format` to the chart resource record..

panaaj:
Can you provide an example of `/resources/charts` output from Artemis so I can compare to @signalk-charts-plugin.

rob42:
/signalk/v1/api/resources looks like this:
```
{
  "urn:mrn:signalk:uuid:22db3599-55a3-40c7-8a7e-4b68dcaac8d0": {
    "identifier": "NZ6142_2",
    "name": "Nelson Harbour and Entrance - Port Nelson",
    "tilemapUrl": "/./mapcache/NZ6142_2",
    "scale": 250000,
    "description": "Nelson Harbour and Entrance - Port Nelson",
    "timestamp": "2018-05-25T22:18:22.448Z",
    "$source": "unknown"
  },
  "urn:mrn:signalk:uuid:e5924681-a697-4f86-9277-b9e2a4ad17db": {
    "identifier": "NZ61403",
    "name": "Mapua",
    "tilemapUrl": "/./mapcache/NZ61403",
    "scale": 20000,
    "description": "Mapua",
    "timestamp": "2018-05-25T22:18:23.015Z",
    "$source": "unknown"
  },
  "urn:mrn:signalk:uuid:3d22463d-09bf-4e3b-b699-715eb9e9e0e6": {
    "identifier": "NZ614",
    "name": "Tasman Bay",
    "tilemapUrl": "/./mapcache/NZ614",
    "scale": 100000,
    "description": "Tasman Bay",
    "timestamp": "2018-07-18T03:50:04.926Z",
    "$source": "unknown"
  },
  "urn:mrn:signalk:uuid:5cff988f-4882-4874-b429-e4a6ec46fc35": {
    "identifier": "NZ614",
    "name": "Tasman Bay",
    "tilemapUrl": "/./mapcache/NZ614",
    "scale": 100000,
    "description": "Tasman Bay",
    "timestamp": "2018-07-18T04:27:24.327Z",
    "$source": "unknown"
  },
.....
```

rob42:
Since there are different formats it makes sense to add a `format` key, based on those in wikipedia.

panaaj:
Here is what the @signalk-charts-plugin provides:
```{"MBTILES_01":{"identifier":"MBTILES_01","name":"MBTILES_01","description":"NORTH ALASKA","bounds":[-202.5,56.1700229829,-112.5,75.4971573189],"minzoom":3,"maxzoom":19,"format":"png","type":"tilelayer","tilemapUrl":"/signalk/v1/api/resources/charts/MBTILES_01/{z}/{x}/{y}","scale":"250000"}}``` 

We will need to work something out or at least outline the minimum mandatory fields / values required for specific chart types to ensure they can be displayed in the correct manner.

**Comment by rob42** [comment]
Interesting. Its a superset with a slightly different interpretation. The extra data is in a file (tilemap.xml) in the dir root on my charts. OL uses that by default but thats not really portable so having it in the resources makes sense

Compared to artemis:
```
{
	"MBTILES_01": {  //artemis uses uri format, since we could get charts from many sources we dont want name collisions
		"identifier": "MBTILES_01",  
		"name": "MBTILES_01",
		"description": "NORTH ALASKA",
		"bounds": [-202.5, 56.1700229829, -112.5, 75.4971573189],  //need to add
		"minzoom": 3,  //need to add
		"maxzoom": 19, //need to add
		"format": "png", //need to add
		"type": "tilelayer", // we should reference a std here, there are lots of formats
		"tilemapUrl": "/signalk/v1/api/resources/charts/MBTILES_01/{z}/{x}/{y}", //artemis returns the dir, not the extension. Probably the extension here makes sense, more control to the chart maker.
		"scale": "250000"
	}
}
```

I will change artemis and my chart processor to suit.

package.json
tkurki: Um, where do we continue this work? What repo, what branch?

panaaj: I think we have evolved beyond a port of an application to more of an inter-operability exercise that is surfacing issues around areas of the spec / server implementations where there are still questions to be answered.

Most of the discussion has been around how the different server implementations differ:
- `/resources/charts` response
- anchor alarm operation
- how to list what applications are installed
The things that an application needs to know to be able to present a "better" user experience.

So ultimately what I think we are getting into here is `what does a well structured /well  behaved Signal K application look like and what does it need from the server?`
- How does it test the server for available capabilities? i.e. history api
- How does it query what applications / plugins are available?
- Is security enabled?
- If it can't PUT values to API paths that needs what should it do? Fail with a message, try an alternative (e.g. send it via an UPDATE message), etc.
- and more......

Happy to take this to a new repo if it makes sense and maybe position the application as a means for testing server api / spec implementations to ensure an application can work across all server types.

Prior to the port there was Freeboard functionality that did not work on node server but mostl likely did *(I have only used the node server so I am making an assumption here)* on the Java / Artemis server. We are now testing the port for similar operation on both server types which is where most of the dialog is.

rob42:
Ive created a new pull request #51 so that this can be continued. Its still based on the panaaj-angular branch, and should include everything to date.

panaaj:
So I have an update ready to push... do you want me to push it to #51 ?

package.json
tkurki: As things did not work perfectly before I don't think we need to resolve all problems before publishing. Your work is a delight and a major step forward I'd like to push this out and take care of the smaller gripes piecemeal afterwards.

One major advantage in publishing this as a new version of Freeboard is that people who have Freeboard already installed will get notified of the available upgrade and can upgrade with one click.

I'll continue the discussion in #51.

rob42:
If you push to panaj-angular it will  end up in #51

PR #52, opened 2018-09-26
chore: build from root to public

tkurki:
These changes allow publishing from the project's root
directory with `npm publish`, which will trigger the `prepare`
script that will run the Angular build with /public as the target
directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/public
+/node_modules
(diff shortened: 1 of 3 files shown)


panaaj:
It looks OK... tested with my local install to confirm `ng build` can output to a parent folder, all good.

PR #59, opened 2018-10-17
[WIP]: Navigate To... waypoints and routes

panaaj:
Add functionality to send requests to Signal K server to enable:

- `Navigate to` Waypoint
- `Activate` Route

Determine what information needs to be sent to the Signal K server to set a route as the `activeRoute` or a waypoint as the `nextPoint` which will then trigger the transmission of the relevant `course` data.

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG: Freeboard-SK
 
+### v1.1.0
+
+- Display Navigation data from `navigation.courseGreatCircle.nextPoint` / `navigation.courseRhumbline.nextPoint`
+
+
+**Breaking Changes:**
+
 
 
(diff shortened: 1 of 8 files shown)


panaaj:
Need to confirm the minimum data required by the server from the client to deem the destination point set.
i.e for a Waypoint is it:
-  `navigation.courseGreatCircle.nextPoint.value= {href: "/resources/waypoint/<uuid>", type: "Waypoint"}`
OR
- `navigation.courseGreatCircle.nextPoint.position= {latitude: -28.45, longitude: 138.654}`

For a Route is it:
-  `navigation.courseGreatCircle.activeRoute.href= "/resources/route/<uuid>"`

panaaj:
Navigate To function included in #64.

PR #60, opened 2018-10-17
Fix: app list url for non-node servers to include api version number.

panaaj:
Update the URL used to retrieve the list of applications from the Signal K server for non Node server types to include the api version number:
`/signalk/v1/apps/list`


panaaj:
Just an FYI the url fix aligns the url with the [Spec Issue #504](https://github.com/SignalK/specification/issues/504).

PR #85, opened 2020-10-03
Animated map updates

Skeen:
This PR:
1. Fixes a typo in README.md
2. Introduces a `docker`-based development environment
3. Enables OpenLayers view animation for dynamic map updates, instead of the jerky direct updates from before.

Considers:
* Maybe the animation should be an option, as to save cpu cycles and power, or for people who simply prefer the old behavior?

Before:
before

After:
after

Note: The low framerate of the gif, does not do the animation justice.


panaaj:
As this PR addresses more than one topic I will respond to each individually.
The preference is to address only one issue / feature per PR.

- Introduces a docker-based development environment
Whilst docker can be a valuable environment in the development process, it is more of a developer preference than a requirement. So I am not convinced that the `Dockerfile` is required although it would be a good FAQ topic for the [wiki page](https://github.com/SignalK/freeboard-sk/wiki/Developing-with-Docker)

- Enables OpenLayers view animation for dynamic map updates, instead of the jerky direct updates from before.
This functionality is added by using the following map component directives:
```
<aol-map 
    [loadTilesWhileAnimating]="true"
    [loadTilesWhileInteracting]="true"....>

    <aol-view  [zoomAnimation]="true" ... >
         ...
    </aol-view>

</aol-map>
```
I also agree that the _map animation_ should be an settings option (which is missing from this PR).
See [animateMap branch](https://github.com/SignalK/freeboard-sk/tree/animateMap) for the proposed changes to support this.

PR #90, opened 2021-01-10
fix: The displayed circle that indicates anchor.maxRadius now has correct diameter

chris0348:
The displayed circle that indicates anchor.maxRadius now has correct diameter. Before setting a radius of e.g. 1852m (1nm) resulted in a displayed circle with a radius of about 1000m. It seems that aol-geometry-circle internally assumes that the radius is given in nautical miles, although the internal units of sk are SI-units (radius [m]). For the fix the argument is scaled by a factor of 1.852.

### DIFF
--- a/src/app/modules/map/fb-map.component.html
+++ b/src/app/modules/map/fb-map.component.html
@@ -430,7 +430,7 @@
     <aol-layer-vector [zIndex]="49" *ngIf="!app.config.anchor.raised">
         <aol-source-vector>
             <aol-feature>
-                <aol-geometry-circle [radius]="app.config.anchor.radius">  
+                <aol-geometry-circle [radius]="app.config.anchor.radius * 1.852">  
                     <aol-coordinate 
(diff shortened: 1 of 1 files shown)


panaaj:
I'm not sure how this issue is presenting but `aol-geometry-cricle` uses the value as an input into `ol.circle.setRadius()` which accepts a number in the units of the projection.
The units of the projection is meters so the number supplied for radis is in meters so should not require any scaling.

I have tested by setting values via the Freebaord anchor watch UI and also directly within the `anchor-alarm` plugin config and _(OL circle surface size rendering accuracy aside)_ the radius of the circle drawn is as expected.

panaaj:
The anchor radius rendering issue is resolved by applying the following formula to caclulate the actual radius on ground at the anchor position... `radiusToDraw= anchor.maxRadius / getPointResolution('EPSG:3857', 1, fromLonLat(position))`.

This fix will be included in the next release.

PR #95, opened 2021-03-24
Add names to the route points when imported from GPX

sergei:
This PR is based on this discussion https://github.com/SignalK/signalk/discussions/22 
Please let me know if my thinking is aligned with your suggestion. 

 I tested it on my setup and It seems to handle both cases:

- Hand drawn routes, with no names  present
- Routes imported from GPX, then points have names. 

Not sure if i covered all route manipulation routines though.


panaaj:
I think I might have beaten you to it in 1.15.2 branch!

sergei:
Very nice! Would be happy to use the proper implementation. 

Thanks  a lot,
  -- Sergei

PR #283, opened 2025-08-04
MOB button

daniel-freiermuth:
This PR adds a prominent MOB button in the bottom-right corner of Freeboard (see screenshot below). It raises a MOB alarm, ~creates a waypoint~ and navigates to it.

When implementing this, I was a bit unsure whether to use MOB or POB as abbreviation. I'd favor POB, but all of the sources except the waypoint type seem to use MOB. I'm happy to align with whatever.

This is my first contribution to the project. I hope you are able to take in this feature. Let me know whether there are any amendments to the PR necessary.

~~Note that this PR also contains two unrelated fixes:~~
- ~~The dev server config in the Readme was outdated~~
- ~~When compiling, I got a severe type error related to click events. The fix was rather easy: using more specific types. But I'm surprised to find type errors on master, so probably something with my setup is off?~~

bild

/cc @panaaj Tagging you as you seem to drive a lot of Freeboard's development

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -743,6 +743,7 @@
           <!-- /right button bar -->
 
           <!-- MOB Emergency Button -->
+          @if(!isInteracting) {
           <div class="mob-emergency-button">
             <button
               class="button-warn mob-button"
@@ -754,6 +755,7 @@
               <mat-icon class="ob" svgIcon="alarm-mob"></mat-icon>
             </button>
           </div>
+          }
           }
           <!--map panel-->
           <div
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -741,6 +741,19 @@
           </div>
           }
           <!-- /right button bar -->
+
+          <!-- MOB Emergency Button -->
+          <div class="mob-emergency-button">
+            <button
+              class="button-warn mob-button"
+              mat-fab
+              (click)="raiseMOBAlarm()"
+              matTooltip="MAN OVERBOARD! Emergency Alarm"
+              matTooltipPosition="above"
+            >
+              <mat-icon class="ob" svgIcon="alarm-mob"></mat-icon>
+            </button>
+          </div>
           }
           <!--map panel-->
           <div
--- a/src/app/app.component.css
+++ b/src/app/app.component.css
@@ -17,7 +17,7 @@ mat-nav-list {
     height: 100vh;
     background-color: black;
     color: white;  
-    overflow: hidden;  
+    overflow: hidden;
 }
 
 .view {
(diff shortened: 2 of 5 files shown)


tkurki:
Buttons are pretty easy to touch by accident - is there a confirmation dialog or long press or smthing to guard against that?

Does it stay or get hidden with the top right "hide extra buttons" button? Should that be configurable? 🤔

daniel-freiermuth:
> Buttons are pretty easy to touch by accident - is there a confirmation dialog or long press or smthing to guard against that?

Not yet. The idea is to take minimal time and attention to raise a MOB alarm. Resetting a false alarm means deleting the alarm, waypoint and reactivating the previous navigation.

Are you thinking about a scenario where the MOB functionality is not desired and would become annoying?

> Does it stay or get hidden with the top right "hide extra buttons" button? Should that be configurable? 🤔

It doesn't get hidden with the top-right "hide extra buttons" button.

Having some configuration options sounds reasonable. I guess not everyone is interested in a MOB button?

panaaj:
Fyi, I am currently doing some work to make the right-side button bar configurable so as to minimise screen clutter.
So the MOB button would be included in this.

panaaj:
Also with regard the automatic creation of a waypoint, is this necessary?
The alarm notification generated by FSK contains the position at which the alarm was raised and this is displayed on the map.
The UI provides the "Go To" action to make this the destination to navigate to.

daniel-freiermuth:
> Fyi, I am currently doing some work to make the right-side button bar configurable so as to minimise screen clutter.
So the MOB button would be included in this.

Sounds good. I don't have strong opinions about where to place the button as long as it is easily accessible. Is there a branch I should base my commits on?

panaaj:
> Is there a branch I should base my commits on?

[branch 2.16-next.1](https://github.com/SignalK/freeboard-sk/tree/2.16-next.1) is the next release candidate.
It contains some re-architecting and prep for revamping settings which will support toolbar customisation.

_PS It already contains the fixes for the changed the Openlayers pointer event types._

daniel-freiermuth:
Nice! I rebased the commits on this branch now.

How do you spawn a test instance? `npm start` doesn't work anymore for me?

panaaj:
ng serve

daniel-freiermuth:
That gives me

```bash
$> npm run ng serve           

> @signalk/freeboard-sk@2.16.0-ng20 ng
> ng serve

Error: Could not find the '@angular-builders/custom-webpack:dev-server' builder's node package.
```

panaaj:
I have the following issues with the current behaviour.....
1. I don't think that having different behaviours depending on the method of raising  MOB is ideal. The result should be the same if the alarm is raised via the Alert List or the button.
2. The action of setting a course to the MOB location should not be tied to the raising of the alarm, but rather an action triggered by the receipt of the notification. This way the action can be configurable and will be consistent regardless of how the alarm was raised (the MOB could be raised in KIP, another APP, hardware button, etc).

src/app/app.component.html
```diff
@@ -741,6 +741,19 @@
           </div>
           }
           <!-- /right button bar -->
+
+          <!-- MOB Emergency Button -->
+          <div class="mob-emergency-button">
```
panaaj, line 746: The button should be hidden when map interaction is active.
Wrap it with `@if(!isInteracting) {...}`

daniel-freiermuth:
> I have the following issues with the current behaviour.....
> 
>     1. I don't think that having different behaviours depending on the method of raising  MOB is ideal. The result should be the same if the alarm is raised via the Alert List or the button.
> 
>     2. The action of setting a course to the MOB location should not be tied to the raising of the alarm, but rather an action triggered by the receipt of the notification. This way the action can be configurable and will be consistent regardless of how the alarm was raised (the MOB could be raised in KIP, another APP, hardware button, etc).

Have you thought along the lines I now implemented?

panaaj:
> I'm wondering whether the MOB-Alarm → Course functionality would be better placed as a separate SignalK plugin now?

If you want the resultant action (set course to the MOB position) to happen when Freeboard is not open on a device... then yes a plugin is where this functionality should go.

Clearly if the MOB button is actioned in Freeboard this is not an issue.

tkurki:
Should this logic be in server? Having a plugin for it seems fragile.

panaaj:
Also will need to have logic to determine what to do when multiple alarms are present.
Clearly it shouldn't just set course for the next alarm that presents.... Or should it?

tkurki:
i was thinking this should be mob specific.

panaaj:
I meant multiple mob alarms.

tkurki:
Just that plugins may or may not be installed and active.

PR #284, opened 2025-08-08
fix: Set minHeight by vh instead of vw

daniel-freiermuth:
The recently pushed version 2.16.1 introduces tabs in the settings page and sets css minHeight to 80vw. This binds the height of the settings dialog to the viewport width and leads to problems on landscape screens.


daniel-freiermuth:
Do you want to take it from here or shall I update the PR? There is also a maxHeigh: 65vh, that doesn't look good.

panaaj:
I've created a new release 2.16.2 with the fix.

PR #298, opened 2025-09-29
feature: config the opacity of online chart layers

night199uk:
Combined with:
https://github.com/SignalK/charts-plugin/pull/39

This will allow you to configure the opacity of a chart when configuring it in the SignalK charts-plugin. This is useful for certain types of layers, such as NOAA weather layers.


tkurki:
I am not sure if having opacity in the charts api is the best solution or cadence. You control what layers are shown and in what order in Freeboard - to me opacity is about the same theme. Opacity by itself, without the order and selection of layers, is out of context.

PR #300, opened 2025-11-05
Add time dimension support for overlay layers

dleone13-terp:
Parse time dimension from WMS/WMTS capabilities and integrate it into the overlay layers.
Update refreshing to consider time dimension when fetching data.
Time dimension is used as an offset for the user, and saved as part of the layer state.


src/app/modules/map/ol/lib/resources/layer-livelayer.component.ts
```diff
@@ -111,22 +115,158 @@ export class FreeboardLiveLayerComponent
     this.clearlayers();
   }
 
+  /** Refresh time dimension bounds from WMS/WMTS capabilities */
+  private async refreshTimeDimension(layerId: string, liveLayer: LiveLayer) {
```
panaaj: Time dimension processing should be happening outside of the live-layer component.
The component's purpose is to render the layer.... dimension changes should be passed to the component.

src/app/modules/map/ol/lib/resources/layer-livelayer.component.ts
```diff
@@ -111,22 +115,158 @@ export class FreeboardLiveLayerComponent
...
-    this.layerMap.forEach((v, k) => {
-      //console.log(k, v.infoLayer.values.refreshInterval, v.lastRefresh, n - v.lastRefresh)
-      if (
-        v.infoLayer.values?.refreshInterval > 0 &&
-        n - v.lastRefresh >= v.infoLayer.values?.refreshInterval
-      ) {
-        // refresh layer source
-        const src = v.layer.getSource();
-        src?.refresh();
+    for (const [k, v] of this.layerMap.entries()) {
```
panaaj: Maintenence of time values should not be in this component.

panaaj:
Generally, I think that loading and rendering of layers should use the default behaviour / time dimension,
and that application of the time dimension is temporary and  it should revert upon reload of the layer(s).

PR #330, opened 2026-02-14
add: TileJSON metadata enrichment, fix tests, and enable CI test runs

herostrat:
TileJSON charts now pull missing metadata [minzoom] [maxzoom] [format] so map zoom/bounds and the info dialog show correct values.

In addition I wanted to add tests for this and found that the test framework was not working correctly and niot hooked up to the CI.
I fixed this. 

I thought about adding the lockfile in order to be able to use npm ci (and than one could cache it in CI). I could add it if you'd like, but I didn't want to add another thing on top in this PR. I can send a PR afterwards with it or add it here - whatever you prefer.

Note: CI fails because it of errors fixed in this PR: https://github.com/SignalK/freeboard-sk/pull/329
A CI can be seen here:  See: https://github.com/herostrat/freeboard-sk/pull/1

A visual before and after image:
image

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -17,9 +17,9 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: npm
 
-      - name: npm ci
+      - name: npm install
         run: |
-          npm ci
+          npm install
 
       - name: Install Chromium
         run: |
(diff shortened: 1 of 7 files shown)


herostrat:
I forgot to add format and the vector layers and fixed it now:
image
image

joelkoz:
Thanks for the contribution.  Can you remove all of the modifications to the CI and focus this on just the TileJSON fix. We've recently redone the CI to match the shared CI used across all the SignalK projects.  If you can update the PR (including title and description) and rebase against the current master I can take another look.

PR #341, opened 2026-04-03
Improve startup performance: eliminate duplicate API calls

### LINKED ISSUE #340: Improve startup performance: eliminate duplicate API calls
## Problem

During startup, `fetchResources(true)` fires twice — once on `config$.ready` (before the server connection is established) and again in `queryAfterConnect()` (after connect). This doubles every resource request on every page load.

Observed in a HAR capture (production build, real device):

| Endpoint | Times called | Transfer per call |
|----------|-------------|-------------------|
| `/v2/api/resources/tracks` | 3x | ~1.6 MB compressed (~8 MB uncompressed) |
| `/v2/api/resources/routes` | 2x | ~12 KB |
| `/v2/api/resources/waypoints` | 2x | ~1.4 KB |
| `/v2/api/resources/charts` | 2x | ~2.2 KB |
| `/v2/api/resources/notes` | 2x | ~0.2 KB |
| `/v2/api/resources/regions` | 2x | |
| `/v2/api/resources/infolayers` | 3x | |
| `skServer/loginStatus` | 4x | |

The tracks endpoint alone accounts for ~4.8 MB of redundant transfer and ~30s of cumulative server time across the 3 calls.

## Root cause

1. `AppFacade` constructor calls `loadConfig()` → `parseLoadedConfig()` → `doPostConfigLoad()` which emits `config$.ready`
2. `AppComponent` constructor subscribes to `config$` and calls `fetchResources(true)` on `ready` — **this is the redundant call**
3. `AppComponent.ngOnInit()` calls `connectSignalKServer()` → `onConnect()` → `queryAfterConnect()` → `fetchResources(true)` — **this is the correct call**

Additionally, there is no deduplication on `listFromServer()`, so concurrent calls for the same resource type each fire a separate HTTP request.

## Proposed fix

- Remove the `fetchResources(true)` call from the `config$.ready` subscription — resources should only be fetched after the server connection is established
- Add in-flight request deduplication to `listFromServer()` so concurrent calls for the same collection return the existing promise

msallin:
## Summary

- Remove the redundant `fetchResources(true)` call from the `config$.ready` subscription — resources are already fetched in `queryAfterConnect()` after the server connection is established
- Add in-flight request deduplication to `listFromServer()` so concurrent calls for the same resource collection return the existing promise instead of firing parallel HTTP requests

Closes #340

## Test plan

- [ ] Verify resources (routes, waypoints, charts, notes, regions, tracks) are still loaded correctly after startup
- [ ] Check browser DevTools Network tab: each `/v2/api/resources/*` endpoint should be called only once during startup, not twice
- [ ] Verify that importing GPX/GeoJSON still triggers a resource refresh

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -273,16 +273,6 @@ export class AppComponent {
     // set self to active vessel
     this.app.data.vessels.active = this.app.data.vessels.self;
 
-    // CONFIG$ - handle app.config$ event
-    this.obsList.push(
-      this.app.config$.subscribe((value: string) => {
-        // config has been loaded and cleaned (ready)
-        if (value === 'ready') {
-          this.fetchResources(true);
-        }
-      })
-    );
-
     // handle skAuthChange signal
     effect(() => {
       this.app.debug('** skAuthChange Event:', this.app.skAuthChange());
--- a/src/app/modules/skresources/resources.service.ts
+++ b/src/app/modules/skresources/resources.service.ts
@@ -77,6 +77,7 @@ export type SKSelection = SKResourceType | 'aisTargets' | 'infolayers';
 @Injectable({ providedIn: 'root' })
 export class SKResourceService {
   private reOpen: { key?: string; value?: string; readOnly?: boolean };
+  private inflight = new Map<string, Promise<any[]>>();
 
   constructor(
     private dialog: MatDialog,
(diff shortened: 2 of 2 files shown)


panaaj:
This PR creates other issues and is more appropriately addressed by tweaking the sequencing to ensure that resources are loaded after config parsing.

I have created a fix to address this accordingly.

PR #345, opened 2026-04-04
fix: stack chart boundary labels to prevent overlap

### LINKED ISSUE #344: Chart boundary labels overlap when multiple boundaries are shown
## Description

When activating "Boundaries" in the Charts list, all chart boundary labels render at the same vertical position, causing them to overlap and become unreadable when multiple charts have similar or overlapping bounds.

Image

## Steps to Reproduce

1. Open the Charts list
2. Activate "Boundaries"
3. Observe that chart boundary labels are rendered on top of each other

## Expected Behavior

Chart boundary labels should be vertically stacked so each label is readable.

## Actual Behavior

Labels overlap at the same position, making them unreadable.

msallin:
## Summary

Fixes #344

- Chart boundary labels now use staggered vertical offsets based on their index, preventing labels from overlapping when multiple boundaries are displayed

## Changes

In `layer-chart-bounds.component.ts`:
- Added `labelIndex` property to each boundary feature
- Calculate `offsetY` as `-10 - (index * 16)` so each label is shifted 16px further up
- Also fixed minor whitespace in `rgba` fill value

### DIFF
--- a/src/app/modules/map/ol/lib/resources/layer-chart-bounds.component.ts
+++ b/src/app/modules/map/ol/lib/resources/layer-chart-bounds.component.ts
@@ -73,6 +73,7 @@ export class ChartBoundsLayerComponent extends FBFeatureLayerComponent {
       ];
       f.set('labelPos', lp);
       f.set('boundColor', this.boundStyles[i]);
+      f.set('labelIndex', i);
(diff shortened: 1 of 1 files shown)


panaaj:
Label decluttering enabled to address this issue in release 2.21.0

PR #352, opened 2026-04-05
feat(charts): add In view filter to chart list

### LINKED ISSUE #351: Feature: filter chart list to charts within current map viewport
## Problem

When working in an area with many charts installed, the chart list quickly becomes long and it's cumbersome to scroll through charts that cover other parts of the world. Currently the only way to filter the list is by typing into the text filter, but this requires remembering each chart's name.

## Proposed feature

Add an **In Bounds** toggle to the chart list. When enabled, the list is restricted to charts whose bounds (`minLon, minLat, maxLon, maxLat`) intersect the current map viewport extent. The list updates reactively as the user pans/zooms the map.

## Behaviour

- Off (default): current behaviour — all charts visible, subject to text filter.
- On: show only charts whose `bounds` intersect `app.mapExtent`. Charts without a `bounds` property (e.g. OSM) are treated as global and remain visible.
- Combines with the existing text filter (both predicates applied).
- Placed next to the existing **Bounds** (display chart boundaries on map) toggle.

## Result

chart list with In Bounds toggle

## Implementation

Draft PR: will be linked below.

Uses `intersects` from `ol/extent` for the bbox test. Reacts to `app.mapExtent` signal changes via an Angular `effect` that only recomputes when the toggle is on.

msallin:
## Summary
- Adds an **In view** slide toggle to the chart list. When enabled, only charts whose `bounds` intersect the current map viewport (`app.mapExtent`) are shown.
- The list updates reactively as the user pans/zooms the map, via a signal-driven `effect` that recomputes the filter while the toggle is on.
- Combines with the existing text filter; charts without bounds metadata are treated as global and remain visible.

## Changes
- `chart-utils.ts`: new pure `isChartInView(bounds, extent)` helper (alongside `extentFromBounds`) encapsulating the bounds/viewport intersection and the "no bounds = global" rule. Both inputs are EPSG:4326 `[minLon, minLat, maxLon, maxLat]`, compared via `intersects` from `ol/extent` with no re-projection.
- `chart-utils.spec.ts`: unit tests for `isChartInView` (overlap, containment, edge-touching, disjoint, missing/malformed bounds) plus boundary tests for `extentFromBounds`.
- `chartlist.ts`: new `inViewOnly` flag, `toggleInViewOnly()`, an overridden `doFilter()` combining the name filter with the viewport filter, and an `effect()` watching `app.mapExtent`.
- `chartlist.html`: new **In view** `mat-slide-toggle`; the existing boundary toggle relabelled **Bounds** and the Re-order / Add buttons made icon-only to fit on one row.

## Known limitation
The intersection is a plain min/max overlap test and does not handle viewports crossing the antimeridian (±180° longitude); this is documented in `isChartInView`.

## Closes
Closes #351

## Test plan
- Install several charts covering different regions (e.g. NOAA US + OSM + a European chart)
- Open chart list, confirm the **In view** toggle appears next to **Bounds**
- Toggle on, pan/zoom the map — list shrinks to charts intersecting the viewport
- Verify OSM (no bounds) stays visible regardless of viewport
- Combine with the text filter — both predicates applied
- Toggle off — full list returns

## Summary by CodeRabbit

* **New Features**
  * Added an “In view” filter to the chart list to show only charts visible in the current map extent.
  * Updated the chart-list toolbar controls to a more compact icon-based layout.

* **Bug Fixes**
  * Improved in-view filtering to update automatically as the map moves.
  * Enhanced bounds edge handling near the dateline and world limits, including correct inclusion/exclusion at exact edge values.

* **Tests**
  * Expanded coverage for chart extent visibility and edge-case scenarios.

### DIFF
--- a/src/app/modules/map/ol/lib/charts/chart-utils.spec.ts
+++ b/src/app/modules/map/ol/lib/charts/chart-utils.spec.ts
@@ -1,5 +1,9 @@
 import { expect, describe, it } from 'vitest';
-import { extentFromBounds, resolveLayerMaxZoom } from './chart-utils';
+import {
+  extentFromBounds,
+  isChartInView,
+  resolveLayerMaxZoom
+} from './chart-utils';
 
 describe('resolveLayerMaxZoom', () => {
   it('returns chart max when over-zoom disabled', () => {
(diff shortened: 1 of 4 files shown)


joelkoz:
Thanks for the contribution!  Are you still working on this (it's marked as a draft).  Let me know...

joelkoz:
Thanks @msallin — this is a clean, well-tested addition, and I really like that you documented the antimeridian limitation right in the JSDoc rather than papering over it. The `isChartInView` test coverage (overlap / containment / edge-touch / disjoint / malformed bounds) is exactly the kind of regression guard I like to see. This is a real quality-of-life win for users running large multi-region chart libraries.

Two things before I merge:

1. **Accessible labels on the toolbar buttons.** Switching Re-order and Add to icon-only `mat-icon-button` to fit the new toggle on one row dropped their accessible names — a `matTooltip` isn't read as a label by assistive tech. Could you add an `aria-label` to both buttons (matching the action)? CodeRabbit flagged the same thing inline.

2. **Antimeridian — your call on scope.** The ±180° gap you documented is fine to land as-is in my book, since the worst case is a chart being mis-listed near the dateline, not a bad line on the chart. If you'd rather close it out here, the seam is right where you noted in `isChartInView` — detect a wrapped extent and split it into two longitude ranges before the `intersects` test, plus a regression test for a dateline-crossing view. Equally happy to take it as a follow-up — just let me know which way you want to go.

Once the button labels are in I'm good to merge. Thanks again for the contribution!

msallin:
Thanks @joelkoz. Added `aria-label`s to both the Re-order and Add buttons. I also went ahead and closed out the antimeridian case: `isChartInView` now splits a dateline-crossing viewport into two normalised longitude ranges before the overlap test, with regression tests for charts on either side of ±180. Ready for another look.

PR #355, opened 2026-04-06
fix: constrain tile layers to chart bounds to prevent 404 floods

msallin:
Uses the chart bounds metadata already provided by Signal K to set the OpenLayers layer extent on all tile layer types. This way OL simply won't request tiles outside the chart's coverage area.

Applies to: XYZ raster, TileJSON, WMS, WMTS, PMTiles (raster + vector), and MVT/PBF vector tile layers. The S-57 vector layer already did this via VectorLayerStyleFactory, so this aligns the rest.

Not tested beyond a type-check. Happy to refine.


panaaj:
A more complete fix has been incorporated into the 2.21.0 release.

PR #357, opened 2026-04-06
perf: skip redundant notes refetch on zoom change

msallin:
Notes are fetched by position + radius (`?position=[lon,lat]&distance=X`), so the
cached set is valid regardless of zoom level. Previously, every zoom change triggered
a new GET request to the server, even zooming in, which only narrows the viewport
within already-fetched data.

This also eliminates a double fetch on startup where `fetchResources()` and the
initial `ngOnChanges(mapZoom)` both called `refreshNotes()`.

Follow-up: viewport-aware caching for pan, skip fetch if all viewport corners fall
within the last-fetched circle, instead of the current center-distance threshold.

### DIFF
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -1701,7 +1701,7 @@ export class FBMapComponent implements OnInit, OnDestroy {
 
   // ** called by onMapMoveEnd() to render features within map extent
   private renderMapContents(zoomChanged?: boolean) {
-    if (this.shouldFetchNotes(zoomChanged)) {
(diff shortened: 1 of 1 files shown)


msallin:
I thought a bit about that PR and Im not happy. I leave it open for discussion but I think we better address that with a more robust approach (i.e. checking abounds). Maybe we can improve that for all resources. Like tracks are currently fetched without any filter.

joelkoz:
Given you are not happy with this PR, I am going to close it.  If your unease is simply this particular solution but you want to keep working on the issue, let's open an issue and discuss it there

Issue #285, opened 2025-08-11
Lookahead


tkurki:
![Image](https://github.com/user-attachments/assets/67b2b762-8a65-49b5-898f-aac21986f0e7)

Usually you are interested more in what lies in the direction of travel rather than what lies behind you. This is not always true, as you may be tacking or you are navigating a course that is taking a zigzag or just on a curving course.

I think Freeboard would be even more useful if it would have an adjustable lookahead so that when following the vessel it would keep the vessel not at the center but so that cog would point at the middle and the boat would be x percent of half the windows’s shorter dimension offset.

Maybe when navigating a course or to a destination it could be even more intelligent, but  cog based offset would be a safe universal solution.

panaaj:
@tkurki @macjl @daniel-freiermuth  [2.17.0-beta.1](https://www.npmjs.com/package/@signalk/freeboard-sk) contains this feature if you want to try it out.

torevalley:
Additional ideas around area of look-ahead and distance rings. 
Skippo a popular navigation app in Sweden has a heading/cog pointer with distance and time on it, see screenshot. 


<img width="1620" height="2160" alt="Image" src="https://github.com/user-attachments/assets/ef7cc117-94ec-4e47-a7af-9eeb3cb19819" />

torevalley:
Yes I know, but instead of setting the length in minutes to decorate the COG/Heading line with distance and time markers is a suggestion.

panaaj:
@torevalley  Can you please create a new issue for this feature so it doesn't get lost.

torevalley:
Done

panaaj:
Available in v2.17.0

PR #359, opened 2026-04-10
fix(lookahead): Center map based on COG instead of north of vessel.

daniel-freiermuth:
Disclaimer: Maybe I missunderstood the lookahead mode introduced in #285.

Currently, the lookahead mode centers the map north of the vessel.

This patch calculates the map center based of the COG since I expected the lookahead mode to use the COG/heading or course to determine the center offset.

Addtionally, this patch uses a different calculation for the base offset distance. Instead of using the tiles' extents (which can be a lot bigger than the viewport's extents if the viewport is rotated) it uses the viewport corners to determine the extends. In head-up mode, this places the vessel nicely in the lower part of the screen. In north-up mode, the vessel moves on a rectangle with same the aspect ratio as the screen.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1323,7 +1323,7 @@ export class AppComponent {
    * Center the map relative to the vessel position
    */
   protected centerVessel() {
-    const pos = this.app.calcMapCenter(this.app.data.vessels.active.position);
+    const pos = this.app.calcMapCenter();
     this.centerAndZoom(pos);
   }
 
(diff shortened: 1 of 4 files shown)


panaaj:
The current functionality is by design... maybe the term look-ahead is a little ambiguous and could be reviewed.

daniel-freiermuth:
Thanks for your response. Could you quickly summarize when and how to use the "look-ahead" feature? I don't understand how to use it.

panaaj:
Apologies, I was a bit premature with my comments (and rushed review). I will merge.

PR #364, opened 2026-04-17
fix(map): draw tracks and routes crossing antimeridian without jumps

daniel-freiermuth:
## Problem

Tracks and routes are drawn using `mapifyCoords` which also handles the drawing across the antimeridian.
   The function had two bugs that caused tracks and routes to "jump" across
   the globe when crossing the antimeridian:

   - Segments with a large longitudinal range were not handled correctly.
   - Each point was compared to the *first* point rather than its predecessor,
     so jumps accumulated on multi-segment tracks.

   Additionally, unwrapping coordinates beyond ±180° (the old fix) caused
   OpenLayers to render features only once instead of cloning them into adjacent
   world copies — so a circum-global track would appear to vanish on one half
   of the screen.

   ## Solution

   `mapifyCoords` is replaced by `splitAtAntimeridian`, which splits each
   crossing segment at exactly ±180°, inserting Mercator-interpolated boundary
   points. All resulting sub-segments stay within [-180°, 180°], so OL's wrapX
   mechanism correctly clones every piece into adjacent world copies.

   ## Scope

   This PR touches three related areas, all of which are required for correct
   antimeridian behaviour:

   **1. Track / trail rendering** (`layer-vessel-trail`, `layer-tracks`,
   `layer-aistargets-track`): replace `mapifyCoords` with `splitAtAntimeridian`,
   change geometry type from `LineString` to `MultiLineString` where needed.

   **2. Route rendering and editing** (`layer-routes`): routes are now
   `MultiLineString` features split at the antimeridian.  Route editing required
   two coupled changes:
   - The OL `Modify` interaction sees a scratch `LineString` built with
     `mapifyCoords` (unwrapped, no split-boundary vertices), so the user never
     drags a phantom ±180° vertex.
   - After each vertex drop, `FreeboardRouteLayerComponent.updateRouteSegments`
     re-splits at the antimeridian and updates the rendered `MultiLineString`.
     The coordinates passed back to `measurementCoords` (and saved to the route)
     are always the original waypoints — the ±180° boundary points are never
     persisted.

   Without the editing changes, @panaaj's concern from the original review
   (phantom vertices becoming saved waypoints) would apply.

   **3. World-copy popup and overlay placement** (`map.component.ts`,
   `fb-map.component.ts`): click events now use `transform()` instead of
   `toLonLat()` so that the raw Mercator x-coordinate (which encodes the world
   copy, e.g. 190° for world +1) is preserved. Vessel / AIS / AtoN popups and
   overlays are shifted to the world copy the user actually clicked — without
   this, clicking a vessel in the eastern world copy would place the popup in
   the primary [-180°, 180°] copy, visually detached from the icon.

   These three areas form a single coherent fix: the rendering change alone
   produces correct visuals, but the editing and popup changes are necessary for
   correct interactive behaviour once features span ±180°.

   ## Testing

   `src/app/modules/map/ol/lib/util.spec.ts` covers `splitAtAntimeridian`:
   eastbound crossing, westbound crossing, large longitude span, multiple
   crossings, Mercator latitude interpolation, and the invariant that ±180°
   boundary points are not present in the original coordinate list (and therefore
   cannot be persisted as route waypoints).

## Some screenshots of the problem and how it looks fixed
 
Broken
Screenshot From 2026-04-16 22-32-09
Fixed
Screenshot From 2026-04-17 08-10-03
Broken:
Screenshot From 2026-04-16 22-33-11

Fixed
Screenshot From 2026-04-17 08-07-22

Route broken:
Screenshot From 2026-04-17 18-42-19
 Route fixed:
Screenshot From 2026-04-17 18-45-56

## Summary

- **Bug Fixes**
  - Improved map rendering and overlay alignment when crossing the international date line (vessels, AIS tracks, routes, and trails).
  - Popovers now consistently open at the selected wrapped “world copy” position.
  - Route editing preserves correct geometry (avoids phantom vertices) and refreshes visible route segments immediately.
  - Route waypoint markers and directional symbols display correctly near ±180°.
  - Map click and right-click coordinates now retain wrapped longitude.

- **Tests**
  - Added coverage for antimeridian splitting and coordinate unwrapping helpers.

### DIFF
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -38,7 +38,10 @@ import {
   S57_CLICKABLE_LAYERS,
   S57_NAMES
 } from './popovers';
-import { FreeboardOpenlayersModule } from 'src/app/modules/map/ol';
+import {
+  FreeboardOpenlayersModule,
+  FreeboardRouteLayerComponent
+} from 'src/app/modules/map/ol';
 import { CoordsPipe } from 'src/app/lib/pipes';
 
 import { computeDestinationPoint, getGreatCircleBearing } from 'geolib';
@@ -213,6 +216,8 @@ export class FBMapComponent implements OnInit, OnDestroy {
 
   @ViewChild(MatMenuTrigger, { static: true }) contextMenu: MatMenuTrigger;
   @ViewChild('olMap', { static: false }) olMap: MapComponent;
+  @ViewChild(FreeboardRouteLayerComponent)
+  routeLayer: FreeboardRouteLayerComponent;
 
   scaleUnits = input<string>('');
 
@@ -1139,8 +1144,14 @@ export class FBMapComponent implements OnInit, OnDestroy {
     }
     let pc;
     if (fid.split('.')[0] === 'route') {
+      // c is a flat Coordinate[] from the scratch LineString (EPSG:3857)
       pc = this.transformCoordsArray(c);
       this.mapInteract.measurementCoords = pc;
+      // Re-split at antimeridian and update all source segment features so
+      // the rendered route reflects the edit after each vertex drop.
+      if (this.routeLayer) {
+        this.routeLayer.updateRouteSegments(fid.split('.')[1], pc);
+      }
     } else if (fid.split('.')[0] === 'region') {
       for (let e = 0; e < c.length; e++) {
         if (this.isCoordsArray(c[e])) {
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -1039,7 +1039,7 @@ export class FBMapComponent implements OnInit, OnDestroy {
         this.mapInteract.draw.features.getArray()[0] as Feature;
       // Give OL Modify a scratch LineString with unwrapped (non-split) coords
       // so there are no phantom ±180° boundary vertices during editing.
-      const mc = mapifyCoords([...rawCoords]);
+      const mc = mapifyCoords(rawCoords);
       // Shift the scratch geometry to the world copy the user clicked in.
       // OL always returns the primary-world Feature regardless of which
       // rendered copy was hit, so we identify the correct copy explicitly.
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -46,13 +46,8 @@ import { CoordsPipe } from 'src/app/lib/pipes';
 
 import { computeDestinationPoint, getGreatCircleBearing } from 'geolib';
 import { fromLonLat, toLonLat } from 'ol/proj';
-import {
-  LineString as OlLineString
-} from 'ol/geom';
-import {
-  fromLonLatArray,
-  mapifyCoords
-} from './ol/lib/util';
+import { LineString as OlLineString } from 'ol/geom';
+import { fromLonLatArray, mapifyCoords } from './ol/lib/util';
 import { Style, Stroke, Fill } from 'ol/style';
 import { Collection, Feature } from 'ol';
 import { Feature as GeoJsonFeature } from 'geojson';
@@ -539,7 +534,8 @@ export class FBMapComponent implements OnInit, OnDestroy {
           if (this.overlay().type === 'ais') {
             this.overlay.update((current) => {
               const newPos = this.dfeat.ais.get(current.id).position;
-              const wo = Math.round((current.position[0] - newPos[0]) / 360) * 360;
+              const wo =
+                Math.round((current.position[0] - newPos[0]) / 360) * 360;
               return Object.assign({}, current, {
                 position: wo !== 0 ? [newPos[0] + wo, newPos[1]] : newPos,
                 vessel: this.dfeat.ais.get(current.id)
@@ -1056,21 +1052,26 @@ export class FBMapComponent implements OnInit, OnDestroy {
       const clickMercX = clickMerc[0];
       const clickMercY = clickMerc[1];
       const worldWidth = 2 * 20037508.3427892;
-      let minX = mercCoords[0][0], maxX = mercCoords[0][0];
+      let minX = mercCoords[0][0],
+        maxX = mercCoords[0][0];
       for (const c of mercCoords) {
         if (c[0] < minX) minX = c[0];
         if (c[0] > maxX) maxX = c[0];
       }
       const m = Math.round(clickMercX / worldWidth);
       const n = Math.ceil((maxX - minX) / worldWidth);
-      let bestShift = 0, bestDist = Infinity;
+      let bestShift = 0,
+        bestDist = Infinity;
       for (let k = m - n - 1; k <= m + n + 1; k++) {
         const shift = k * worldWidth;
         for (const c of mercCoords) {
           const dx = c[0] + shift - clickMercX;
           const dy = c[1] - clickMercY;
           const dist = dx * dx + dy * dy; // squared Euclidean, no sqrt needed
-          if (dist < bestDist) { bestDist = dist; bestShift = shift; }
+          if (dist < bestDist) {
+            bestDist = dist;
+            bestShift = shift;
+          }
         }
       }
       if (bestShift !== 0) {
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -518,8 +518,10 @@ export class FBMapComponent implements OnInit, OnDestroy {
     ) {
       if (this.overlay().isSelf) {
         this.overlay.update((current) => {
+          const newPos = this.dfeat.self.position;
+          const wo = Math.round((current.position[0] - newPos[0]) / 360) * 360;
           return Object.assign({}, current, {
-            position: this.dfeat.self.position,
+            position: wo !== 0 ? [newPos[0] + wo, newPos[1]] : newPos,
             vessel: this.dfeat.self
           });
         });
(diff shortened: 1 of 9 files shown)


panaaj:
What happens when you edit (move points) of a route that crosses the anti-meridian?
Splitting to render in this fashion adds points to the feature which may result in additional (unwanted points) in the route.
This was the initial reasoning for the current approach.

daniel-freiermuth:
Good catch. I think I found a solution, but need to review it once again to make sure I didn't miss anything.

daniel-freiermuth:
> What happens when you edit (move points) of a route that crosses the anti-meridian?
Splitting to render in this fashion adds points to the feature which may result in additional (unwanted points) in the route.

This PR proses now to build editing routes around moving waypoints. For making this happen, this PR customizes the editing interaction. The inserted support points on the anti-meridian will not appear as draggable waypoints. After each moved waypoint, the resulting route is recalculated. Moving a waypoint might lead to a segment "switching direction" and going east instead of west and vice versa if the direction of the shortest path changed. I couldn't make this recalculation happen while dragging, but only when the drag finished. This meant that the intermediate dragged geometry as drawn by openlayers would be misleading (and also expose the support points on the antimeridian if the user changes the segment crossing it). For this reason, I propose not to show the misleading line while dragging, but just update the drawn route whenever a waypoint was updated. Not ideal, but it works. As a bonus, it paves the way to drawing routes segments as great-circle connections (separate PR comes soon).

daniel-freiermuth:
The commits:
- 7c9ebfcc feat: Add splitAtAntimeridian utility function 
- ba94be2d fix: Properly draw **vessel trails** crossing the antimeridian
- 77ad8b89 fix: Split line using rhumb line instead of linear interpolation
- 4b902b7b fix: Properly draw **resource tracks** crossing the antimeridian
- cf1b0986 fix: Properly draw **AIS tracks** crossing the antimeridian
- e438426b refactor: Replace zone-heuristic mapifyCoords with longitude unwrapping. Used while route editing.
- 66455faa fix: Preserve world-copy longitude offset in click events. Make route popups appear in the right world copy.
- 5d4a4979 fix: Place route editing handle in the world copy the user clicked in.
- 75fcb88d feat: Render **routes** as MultiLineString split at the antimeridian
- 6c211356 fix: Show vessel popover in the world-copy the user clicked in

daniel-freiermuth:
Apologies for the wall of commits and PRs. Please let me know whether you're actually interested in the proposed fixed and features and if you'd prefer them in a different format.

joelkoz:
Thanks for working through this. The screenshots make the underlying problem really clear, and I agree this is a real correctness issue for dateline / Pacific / offshore users rather than just a visual nicety.

Before this can land, I'd like to tighten up a couple of things:

1. The branch currently has merge conflicts with `master`, so it will need a rebase/update first.

2. Could you add focused regression coverage for the antimeridian behavior? The most useful seam looks like the coordinate utility logic: eastbound and westbound crossings, large longitude spans, multiple crossings, and ensuring inserted ±180 support points do not become saved route waypoints.

3. The PR has grown beyond rendering tracks/routes: it now also changes route editing, click world-copy handling, and popup placement. Those may all be necessary for the antimeridian fix, but please either split the interaction/world-copy pieces into a separate PR or update the description to make the dependency between them explicit.

4. Per the current contribution convention, please retitle this as something like:
   `fix(map): draw tracks and routes crossing antimeridian without jumps`

I'm interested in getting this fix in; I just want the route-editing invariants and scope to be clear before merge.

daniel-freiermuth:
@joelkoz, thanks for your patience and looking into my PRs. Now I've had some time to take care of it and decided to start with this PR. Let me know if there is anything more that I should adapt.

joelkoz:
@coderabbitai review

joelkoz:
Thanks for sticking with this.

As I was reviewing, the route editing behavior gave me pause. Route editing was reworked in v3.0, after you originally opened this — and in your latest version, dragging a waypoint moves the vertex handle but leaves the line itself sitting still until the drag ends. You'd already flagged that yourself as "not ideal, but it works." Against the pre-3.0 code that was a fair trade; against what's in master now it's a step backwards for everyone who edits a route, dateline or not.

Rather than just push back, I wanted to understand the constraint, so I spent time in the OpenLayers internals (with Claude Opus 4.8). The thing that clicked was the range: OL clones geometry into adjacent world copies not just within one world width but out to ±540° — the canvas renderer re-executes the whole replay group per world (`renderer/canvas/VectorLayer.js` → `renderWorlds`), and `prepareFrame` widens its query with the comment "To support geometries in a coordinate range from -540° to +540°, we add at least 1 world width on each side." So longitudes running continuously past ±180° are a supported input; only geometry beyond ±540° — a track that laps the globe — actually fails.

That matters because your own `mapifyCoords` rewrite already does the right thing: predecessor-relative unwrapping, first point normalised into range, zone heuristic gone. You'd diagnosed the coords[0] and large-span bugs in your description and fixed them in the code. I pulled that rewrite out, kept your regression tests, and merged it on its own as #570 so it isn't entangled with the route work — it's your fix, just landed separately.

What the OL range added on top is only this: with that unwrapping in place, rendering doesn't need to split below ±540°, which covers every realistic route and track. So the split turns out to be unnecessary for rendering — not because your fix was wrong, but because it was already enough once it wasn't fighting the old broken mapifyCoords.

joelkoz:
Following up — I built and ran master with your route, and it turned up something bigger than either of us was looking at.

**The rendering fix (#570) works.** I drew a route from Australia east across the Tasman and out into the Pacific, waypoints well past the dateline (~160°W). It renders as one continuous line, no jump. So #570 alone handles the rendering — your two "broken" screenshots should come good on a rebase, and if you can confirm that on your setup, the rendering question is closed. It also covers tracks, trails and AIS tracks, since those all go through `mapifyCoords` too.

(On the ±540° concern that motivated splitting: it's much harder to reach than it looks. Because the first point normalises into [−180, 180], a single circumnavigation always renders — I confirmed with a Perth→Perth loop. You'd need one unbroken line sweeping ~2 laps in the same direction to break it, and tracks are segmented so any gap resets it. So `splitAtAntimeridian` is optional insurance, not something real usage needs.)

**But you uncovered a foundational bug, and it's much wider than the antimeridian.** While testing I clicked a waypoint east of the dateline and got no popup — until I panned one world west and found it there. Then I checked a route in the Florida Keys, nowhere near the dateline: same thing. Pan one world over, click a feature, and the popup opens in the copy you came from.

The cause is exactly what the "the popup/overlay portion" of your PR is working around: click coordinates get normalised to [−180, 180] on the way in (`toLonLat`), which throws away which world copy the user was looking at, and the overlay is rebuilt in the primary world one width away. It affects every feature type, everywhere — it's in the shipped 3.0.1, not something #570 introduced. And it can't be avoided by constraining the view; OL leaves horizontal panning unbounded for a global projection. I've written it up as issue #572, with the root cause and credit to you.

**What this means for the work here.** With rendering handled at the source (your relevant work moved to #570) and the world-copy bug identified (#572), the route-editing machinery in this PR stops being necessary — it exists to work around exactly those two things:

- rendering: handled by #570
- the vertex-grab-in-the-wrong-world that your `bestShift` search solves: that's #572, the same bug as the popup, just on the hit-test side
- the live line during a drag: returns on its own, because `Modify` edits the real rendered geometry as it did pre-3.0
- saving: already safe — coordinates are normalised on save

So the scratch feature, `updateRouteSegments`, the `@ViewChild` and the `MultiLineString` split all fall away. That's not a knock on the work — it solved real symptoms — it's that with the root causes fixed, the workarounds aren't needed.

**So the one remaining fix that actually matters is #572 — and much of your code here already is it.** The `transform()` change in `map.component.ts` that preserves the world offset, `wrapPos()` for stored-position popups, the overlay-follows-vessel pin, and the `bestShift` edit hit-test are all part of a #572 fix. The one thing I'd change is the boundary: keep the stored/`lonlat` value normalised and apply the offset in Mercator, so a wrapped longitude never exists in degrees and can't reach the server — anything written to Signal K has to stay in [−180, 180] or it breaks other apps reading the data.

Since #570 already delivered what this PR set out to do, I think the cleanest path is to close this one and land #572 as its own focused PR, reusing your world-copy code above. So it comes down to one question: **do you want to own #572, or would you rather I take it?** (And your call on the vehicle — a fresh branch is cleaner, but if you'd rather reshape this one, that works too.)

daniel-freiermuth:
Hi! Thanks for the thorough assessment. Happy for you to take #572.

I can confirm the rendering fix (#570) works — routes crossing the antimeridian display correctly on current master.

On the route editing: the reason I suppressed the live line and rebuilt after each vertex drop was that OL's native preview can show the route going the wrong way around the globe when you move a waypoint. I just verified on current master that this is
 still the case: create a route Paris → Melbourne → LA going eastward, then drag Melbourne to Capetown. The displayed route goes the wrong way around (still eastward) and stays that way until you save.
 
I've moved on to my own chart plotter for my own sailing, so I'll leave it to you whether this is worth addressing. Happy to dig deeper if it's useful — just let me know.

PS: Small bugreport on the side, it seems that now right-clicking deletes the last-touched waypoint, not the waypoint that the right-click is actually performed on.

joelkoz:
Closing this out — but as *superseded*, not declined. Everything this PR set out to do has now landed, in the decomposed form we discussed:

- **Rendering across the antimeridian** — your `mapifyCoords` rewrite went in as #570 (credited to you; the fix and its tests are yours). That alone fixes the track/route "jump", which you confirmed on master.
- **World-copy popups and route editing** — became #572, now implemented in #576. Your work drove all of it: you uncovered the foundational bug, and your `map.component` `transform()` change, `wrapPos()`, and the `bestShift` idea are exactly the shape of the fix. The final version aligns the edit feature by its extent centre (same insight as `bestShift`, arrived at from reproducing your antimeridian route), and it's been verified end-to-end — including a route created two worlds out saving fully in range.

With correct unwrapping (#570) and world-copy awareness (#576) both handled at the source, the route-splitting machinery here isn't needed, so there's nothing left in this branch to merge. But it's the origin of two shipped fixes and a bug that would otherwise still be lurking for every user who pans a world over — that's a real contribution, and thank you for the patience it took to get there.

Thanks again, @daniel-freiermuth.

daniel-freiermuth:
Happy to see Freeboard becoming more robust.

I also saw that you opened #575 for the deletion bug I saw and #574 for the outstanding issue and reason for not drawing the live line.

One more question out of curiosity: Splitting the route was necessary for me because OL did render a geometry in every world-copy only if the geometry stayed within [-180, 180], otherwise it would only be rendered once at the given coordinates. How did you work around that limitation? Was there an update to OL?

joelkoz:
@daniel-freiermuth - honestly not sure if things now work because of an OL update or something else.  My suspicion: the real solution had to do with being very explicit about "these are rendering coordinates, and these are data coordinates saved on the server."  One of the Code Rabbit comments on your original submission had to do with the `mapifyCoords` function mutating parameters, and your fix to satisfy CR was to make a local copy.  Claude said there were three bugs in `mapifyCoords`, and you explicitly fixed two of them with your code, and the third with that Code Rabbit review.  All I know is Claude said the final `mapifyCoords` method (by time we got to it) should work on its own, so it tested it, and it worked.  So, perhaps you were misled by the fact that `mapifyCoords` was the real problem and not OL?

In any event - most of the antimeridian stuff has been fixed, including #575 (I spent several hours testing and iterating with Claude Opus 4.8).  Can you know take a look at #574 from the perspective of the current code as it sits in master and let me know if there is still something to "fix"?  Basically, I can reproduce the bug my making 3 waypoints, but in practice, would you not drop an additional anchor waypoint or two?  Let's discuss over there.

PR #365, opened 2026-04-18
feat: Great Circle predictor

daniel-freiermuth:
This PR builts on top of #361 (and transitively #363) and #364.

This PR adds a Great Circle predictor to the vessel lines. It shows what lies "straight ahead" in a circle around the globe.

Screenshot From 2026-04-18 14-52-08

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -168,13 +168,13 @@ export function cleanConfig(
       laylines: false,
       selfLines: {
         cog: {
-          length: 10, // (minutes) length = cogLine * sog
+          lineLength: { kind: 'time', value: 10 },
           color: 'rgba(204, 12, 225, 0.7)',
           weight: 1,
           dash: 'none'
         },
         heading: {
-          length: -1, // mode for display of heading line -1 = default
+          lineLength: { kind: 'distance', value: 1 },
           color: 'rgba(221, 99, 0, 0.5)',
           weight: 4,
           dash: 'none'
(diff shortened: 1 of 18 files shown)


chris0348:
Interesting new feature! Do you think, you can apply the concept of the great-circle curve to the lines of the measurement module as well? When measuring, at the moment straight lines are shown. Having the great-circle-curve there as well would be great.

panaaj:
If this PR seeks to implement a great-circle predictor for the vessel's current path then let's only include the required changes required to enact that.

daniel-freiermuth:
> If this PR seeks to implement a great-circle predictor for the vessel's current path then let's only include the required changes required to enact that.

Do you want me to change the PR? The idea behind stacking this PR on top of #361 is to avoid merge conflicts and build the great-circle predictor with flexible linestyles (and because #361 is more important to me personally). If you'd prefer to sort out the conflicts later, I can separate the PRs. #363 got already merged. The dependency to #364 got already removed.

Also feel free to do small amendments on this PR's branch if you want. You should have writing access.

Having said this, this PR brings two commits on top of #361:
- 822bb5 Introducing the splitAtAntimeridian function that is necessary for correct drawing of predictors and tracks across the dateline. This commit is split out of #364 and the only overlap left with that PR.
- 4b0e70 Introducing the great-circle predictor

daniel-freiermuth:
> Interesting new feature! Do you think, you can apply the concept of the great-circle curve to the lines of the measurement module as well? When measuring, at the moment straight lines are shown. Having the great-circle-curve there as well would be great.

Yes, I think that should be possible. But it depends on this project's plan whether great-circles are interesting. I figured this PR is a first test for that.

On a side note, I also started building [an experimental chart plotter](https://www.npmjs.com/package/winga-chart-plotter-signalk?activeTab=readme) with a slightly different approach (MapLibre) and goal (web + native) where great-circle is the default.

joelkoz:
Can you fix the merge conflicts when you have time so I can have a closer look?

daniel-freiermuth:
@joelkoz, thanks for looking into this and the other PRs! I'll update them as soon as I have time.

Issue #389, opened 2026-06-01
feature: Custom Freeboard extensions via SK plugin mechanism


joelkoz:
I’d like to propose a lightweight extension mechanism for Freeboard-SK that would allow optional Freeboard enhancements to be installed via normal Signal K server plugins, without requiring users to fork Freeboard or maintain custom builds.

The motivation is that there are several useful customizations that seem valuable but may not belong in the core Freeboard UI for every user. Examples include:

- registering additional icons for notes, regions, or other resources
- registering additional OpenLayers/map styles for custom regions or other map elements
- adding an optional toolbar/menu button
- opening a custom web panel served by the Signal K plugin
- allowing that panel to interact with Freeboard through a small, controlled host API

The goal would be to let the community build optional Freeboard enhancements while keeping Freeboard itself official, upgradeable, and compatible with upstream releases.

## Proposed Discovery Mechanism
Signal K’s Resources API already supports custom resource types through resource providers. Rather than adding a completely separate discovery API, Freeboard could look for a custom resource collection such as:

```
/signalk/v2/api/resources/freeboardExtensions
```

A Signal K plugin that wants to extend Freeboard would register itself as a resource provider for the freeboardExtensions resource type. Its resource entry would be a JSON manifest describing the extension points it contributes.

Example response:
```
{
  "example-extension-provider": {
    "name": "Example Freeboard Extension",
    "description": "Example extension manifest for adding icons, map styles, buttons, and panels to Freeboard.",
    "version": "1.0.0",
    "freeboardApi": "1",
    "icons": [
      {
        "id": "example-custom-icon",
        "url": "/plugins/example-extension-provider/custom-icon.svg"
      }
    ],
    "stylesheets": [
      "/plugins/example-extension-provider/freeboard-extension.css"
    ],
    "mapStyles": [
      {
        "id": "example.customRegion",
        "title": "Example Custom Region",
        "geometry": ["Polygon", "MultiPolygon"],
        "style": {
          "fill": "rgba(0, 120, 220, 0.18)",
          "stroke": "rgba(0, 120, 220, 0.9)",
          "width": 2,
          "lineDash": [6, 4]
        }
      }
    ],
    "buttons": [
      {
        "id": "example-action-button",
        "slot": "mapToolbar",
        "icon": "example-custom-icon",
        "title": "Example New Feature",
        "action": {
          "type": "openPanel",
          "panel": "example-panel"
        }
      }
    ],
    "panels": [
      {
[...]

PR #394, opened 2026-06-09
feat(charts): Custom map symbols via symbol providers

joelkoz:
# Add support for custom map symbols in Freeboard-SK

## Summary

This change lets Freeboard-SK display **custom map symbols** supplied by a Signal
K server, in place of or in addition to its built-in icons. Boaters can give
notes, waypoints, vessels, aids to navigation, and route markers their own
artwork - a custom marker for a waypoint or chart note that is personally meaningful to the user,
a custom boat icon for their own vessel, branded buoys, and so on - without rebuilding Freeboard.

Custom symbols are discovered from the server at startup, offered in the note and
waypoint icon pickers, and rendered on the chart. The change also adds symbol
awareness to **GPX import and export** so symbols survive a round trip through
GPX files.

## Requirements

Custom symbols require a **symbol provider plugin** installed on the Signal K
server. Freeboard consumes whatever symbols the server publishes; it does not
create or store them itself. With no provider installed, Freeboard behaves
exactly as before, using only its built-in icons.

The reference symbol provider is [signalk-symbol-manager](https://www.npmjs.com/package/signalk-symbol-manager), which lets you
upload, edit, and organize symbols through its own web app. Any plugin that
publishes symbols in the standard way will work.

## How overriding works

In addition to adding entirely new note and waypoint markers to Freeport, the user can now 
replace an existing marker icon with one of their own liking by "overriding" it.

Each symbol has a short **id** (for example `dive-site`) and belongs to a
**namespace** (for example your personal `custom` namespace), giving every symbol a
unique name like `custom:dive-site`.

There are two ways a symbol takes effect:

1. **Replace a built-in by name.** Create a symbol whose id matches one of
   Freeboard's built-in icons (for example `dive-site`) ***AND*** has the namespace `fsk:`.
   Freeboard then shows your version everywhere that icon is used — on the chart and in the
   pickers — automatically. You don't have to re-edit existing notes or
   waypoints; anything that already used that built-in picks up your version.

2. **Add a brand-new symbol.** Create a symbol with the namespace `custom` and
   a new id and it appears as an additional choice in the icon pickers (subject to its declared usage — see
   below).

When a custom version of a built-in exists, the pickers show **your version in
place of the built-in**, so you see one entry, not two. A **"Show all symbols"**
toggle reveals everything at once — every custom symbol plus the original
built-ins — so you can still choose the original built-in even after you've
created a custom replacement. If you explicitly pick the built-in this way,
Freeboard remembers that choice and won't silently swap in a custom version
later.

Symbols can advertise which contexts they're meant for (for example "note" or
"waypoint"). Freeboard uses that to keep each picker tidy — the note picker
offers note symbols, the waypoint picker offers waypoint symbols — while "Show
all symbols" ignores the filter and offers everything.

## Where custom symbols can be used

| Area | Custom symbols? |
|------|-----------------|
| Note markers | Yes |
| Waypoint markers (the **Waypoints** category) | Yes |
| Own vessel | Yes |
| AIS vessels | Yes |
| Aids to Navigation (AtoNs) | Yes |
| Route start / turn / end markers | Yes |
| Specialized waypoint types (start pin, start boat, sightings, alarms, pseudo-AtoN) | No — keep their built-in icons |
| Region area shading | No |
| Moored-vessel dots | No |

Markers that rotate to a heading or bearing (vessels, AtoNs, route turn markers,
your own vessel) keep rotating correctly when shown with a custom symbol.

## Overridable built-in symbols and their ids

To replace a built-in, create a custom symbol using the namespace `fsk:` and and id
that matches on of FSK's built in icon Ids:

**Note / waypoint icons** (used for notes and as waypoint icons):

```
anchorage   boatramp   bridge      business    dam
dive-site   ferry      hazard      inlet       lock
marina      dock       turning-basin           radio-call-point
transhipment-dock      notice-to-mariners      diver-down
navigation-structure   fuel        tunnel      waterway-guage
```

**Generic waypoint marker:** `waypoint`

**Own vessel:** `vessel-self`

**AIS vessel icons** (by ship class / state):

```
ais_active   ais_highspeed   ais_special   ais_passenger   ais_cargo
ais_tanker   ais_other       ais_inactive  ais_buddy       ais_self
```

(`ais_self` is the *focused* AIS target, not your own boat — that is
`vessel-self`.)

**Aids to Navigation** — each exists as a *real* (physical) and *virtual*
(AIS-only) form, named `real-…` and `virtual-…`:

```
real-north / virtual-north            real-east / virtual-east
real-south / virtual-south            real-west / virtual-west
real-port / virtual-port              real-starboard / virtual-starboard
real-danger / virtual-danger          real-safe / virtual-safe
real-special / virtual-special        real-basestation / virtual-basestation
real-weatherStation / virtual-weatherStation
real-aton / virtual-aton
```

**Route markers:**

```
route-start      route-waypoint      route-end
```

The route markers and the own-vessel, AIS, and AtoN icons are chosen
automatically by Freeboard rather than picked per feature, so they are overridden
globally: create a symbol with the matching id and every marker of that kind uses
it.

## GPX import and export

Symbols can declare the GPX symbol name (the GPX `<sym>` element) and GPX type
(the GPX `<type>` element) they correspond to. Freeboard uses these to preserve
symbol choices across GPX files.

**Import.** For each waypoint in an imported GPX file that carries a symbol name
(`<sym>`), Freeboard tries to match it to one of your symbols, in order:

1. the symbol whose declared GPX symbol name matches the file's `<sym>`; failing
   that,
2. the symbol whose id matches the file's `<sym>`.

If either matches, that symbol becomes the imported waypoint's icon. If neither
matches, the waypoint gets the default waypoint marker, as before. Matching is
exact and case-sensitive, and uses only the `<sym>` value.

**Export.** When a waypoint that uses a custom symbol is exported to GPX,
Freeboard writes that symbol's declared GPX type and GPX symbol name into the
file's `<type>` and `<sym>` elements (whichever the symbol provides). Combined
with import, this lets a symbol survive a full round trip: export to GPX, re-import,
and the same symbol is restored.

## Behavior without a provider

If no symbol provider is installed, or the server publishes no symbols:

- All pickers and the chart use Freeboard's built-in icons, unchanged.
- The "Show all symbols" toggle simply shows the built-ins.
- GPX import falls back to default markers; GPX export writes whatever the
  waypoint already carried.

There is no configuration to enable — Freeboard uses custom symbols automatically
when they are available.

## Summary by CodeRabbit

* **New Features**
  * Added external symbol/icon resolution across the app, including dynamic rendering for routes, and improved support for vessel self markers.
  * Expanded note and waypoint icon pickers with a “Show all symbols” option to reveal additional selections.
  * GPX import/export now preserves and maps waypoint symbol/icon information more accurately.

* **Bug Fixes**
  * Symbol loading now happens before server data is displayed, reducing missing icons.
  * Improved icon fallback and override behavior for notes/waypoints/routes, including safer handling of unresolved symbols.
  * Improved map marker rotation so symbols stay aligned with the view.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -109,6 +109,7 @@ import {
   SKResourceType,
   WaypointPanel
 } from './modules/skresources';
+import { SymbolService, setSymbolRegistry } from './modules/icons';
 
 interface DrawEndEvent {
   coordinates: LineString | Position | Polygon;
@@ -263,6 +264,7 @@ export class AppComponent {
   private settings = inject(SettingsFacade);
   protected autopilot = inject(AutopilotService);
   protected radarApi = inject(RadarAPIService);
+  private symbols = inject(SymbolService);
 
   constructor() {
     // set self to active vessel
@@ -769,7 +771,7 @@ export class AppComponent {
               }
             })
             .finally(() => {
-              this.fetchResources(true); // fetch all resource types from server
+              this.loadSymbolsThenFetchResources();
             });
           this.getFeatures();
           this.app.data.server = this.signalk.server.info;
@@ -1868,6 +1870,15 @@ export class AppComponent {
   }
 
   // ******** SIGNAL K STREAM *************
+
+  /** Load external symbols then fetch all resource types. */
+  private async loadSymbolsThenFetchResources(): Promise<void> {
+    await this.symbols.load();
+    // Register SymbolService with the module-level hook so pure functions
+    // in app.icons.ts can resolve external symbols without DI injection.
+    setSymbolRegistry(this.symbols);
+    this.fetchResources(true);
+  }
 
   /** fetch resource types from server */
   private fetchResources(allTypes = false) {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1316,7 +1316,7 @@ export class AppComponent {
               this.app.persistToken(r['token']);
               this.app.loadUserConfigfromServer().then((loaded: boolean) => {
                 if (loaded) {
-                  this.fetchResources(true);
+                  this.loadSymbolsThenFetchResources();
                 }
               });
               if (onConnect) {
--- a/src/app/modules/gpx/gpxload-dialog.ts
+++ b/src/app/modules/gpx/gpxload-dialog.ts
@@ -18,6 +18,7 @@ import { MatProgressBarModule } from '@angular/material/progress-bar';
 import { AppFacade } from 'src/app/app.facade';
 import { SignalKClient } from 'signalk-client-angular';
 import { SKResourceService, SKTrack } from 'src/app/modules';
+import { SymbolService } from 'src/app/modules/icons';
 import {
   GPX,
   GPXRoute,
@@ -86,6 +87,7 @@ export class GPXImportDialog implements OnInit {
   protected dialogRef = inject(MatDialogRef<GPXImportDialog>);
   private skres = inject(SKResourceService);
   private signalk = inject(SignalKClient);
+  private symbols = inject(SymbolService);
 
   constructor(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -403,7 +405,7 @@ export class GPXImportDialog implements OnInit {
           this.subCount--;
           if (Math.floor(r['statusCode'] / 100) === 2) {
             this.app.debug('SUCCESS: Route added.');
-            this.app.config.selections.routes.push(skObj[0]);
+            this.skres.selectionAdd('routes', skObj[0]);
           } else {
             this.errorCount++;
           }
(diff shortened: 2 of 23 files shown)


joelkoz:
Modified the symbol provider spec to allow any given symbol to have one or more "aliases" (e.g. multiple sets of `namespace:id` pairs.  Freeboard now only recognizes two namespaces: `fsk:` and `custom:`  If the namespace is `fsk:`, it is assumed the user wants to replace a built in FSK icon with their customized version. If the namespace is `custom:`, then the symbol is presented as an additional symbol the user can choose from to mark a note or waypoint.

joelkoz:
Part of implementation of #389

joelkoz:
@coderabbitai full review

PR #413, opened 2026-06-24
feat(weather): add wind and ocean current overlays

Tominouu:
Adds wind and ocean-current overlays you can toggle from a new Weather panel in the resource sidebar. Both layers are opt-in and off by default.                                                                                                                                                                        

What problem does this solve? 
                                                                                                                                                         
Sailors and cruisers on Signal K could not see live wind or current data on their chart, making passage planning and on-the-fly routing                                                
decisions harder. This PR adds two toggleable vector overlays — wind arrows (from the SK Weather API) and ocean-current arrows (from Open-                                             
Meteo as a documented stopgap) — so a skipper can glance at the chart and immediately understand what the water and air are doing around the boat.                                                                                                                                                                              
                                                                                                                                                                                     
How does it work? 
                                                                                                                                                                
Wind data is fetched via /weather/observations/point on the Signal K server. Ocean currents use the free Open-Meteo Marine API (a stopgap                                              
until SK providers expose currents). Each overlay re-samples the viewport on pan/zoom with a 400ms debounce so a burst of pans collapses                                               
into a single fetch. The Weather panel appears in the resource sidebar only when the server advertises the Weather API (feature flag).                                                 
Toggling the checkboxes shows or hides the overlay; the setting persists across sessions via InfoService.                                                                              

How was this tested?   
                                                                                                                                                                
Manually exercised against a Signal K server with the Weather API enabled. Wind arrows appear in the correct location with correct                                                     
direction/speed labels; panning triggers debounced re-fetch. Currents fall back gracefully when the Open-Meteo endpoint is unreachable.                                                
Config persistence verified via localStorage.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -27,6 +27,10 @@
         <mat-icon>map</mat-icon>
         &nbsp;Charts
       </button>
+      <button mat-menu-item (click)="displayLeftMenu('weatherList', true)">
+        <mat-icon>cloud</mat-icon>
+        &nbsp;Weather
+      </button>
       @if (app.featureFlags().resourceGroups) {
         <button mat-menu-item (click)="displayLeftMenu('resourceGroups', true)">
           <mat-icon>category</mat-icon>
@@ -710,6 +714,11 @@
                     (closed)="displayLeftMenu()"
                   >
                   </anchor-watch>
+                }
+                @if (leftMenuCtrl().weatherList) {
+                  @defer {
+                    <weather-list (closed)="displayLeftMenu()"></weather-list>
+                  }
                 }
               </div>
             }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -79,7 +79,8 @@ import {
   BuildRouteComponent,
   NotePanel,
   RegionPanel,
-  RadarPanel
+  RadarPanel,
+  WeatherListComponent
 } from 'src/app/modules';
 
 import { Convert } from 'src/app/lib/convert';
@@ -173,7 +174,8 @@ interface DrawEndEvent {
     RadarPanel,
     PlotterExtensionOverlay,
     PlotterBackgroundHost,
-    PlotterPanelDrawer
+    PlotterPanelDrawer,
+    WeatherListComponent
   ]
 })
 export class AppComponent {
@@ -205,6 +207,7 @@ export class AppComponent {
     resourceGroups: boolean;
     infoLayerList: boolean;
     anchorWatch: boolean;
+    weatherList: boolean;
   }>({
     leftMenuPanel: false,
     routeList: false,
@@ -216,7 +219,8 @@ export class AppComponent {
     aisList: false,
     resourceGroups: false,
     infoLayerList: false,
-    anchorWatch: false
+    anchorWatch: false,
+    weatherList: false
   });
 
   protected displayFullscreen = signal<{
@@ -1084,7 +1088,8 @@ export class AppComponent {
       aisList: false,
       resourceGroups: false,
       infoLayerList: false,
-      anchorWatch: false
+      anchorWatch: false,
+      weatherList: false
     };
     switch (menulist) {
       case 'routeList':
@@ -1116,6 +1121,9 @@ export class AppComponent {
         break;
       case 'infoLayerList':
         lm.infoLayerList = show;
+        break;
+      case 'weatherList':
+        lm.weatherList = show;
         break;
       default:
         lm.leftMenuPanel = false;
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -358,7 +358,9 @@ export function cleanConfig(
       aisFilterByShipType: false,
       aisState: [],
       resourceSets: {},
-      infolayers: null
+      infolayers: null,
+      weatherWindEnabled: false,
+      oceanCurrentsEnabled: false
     };
   }
 
@@ -391,6 +393,13 @@ export function cleanConfig(
   if (typeof settings.selections.regions === 'undefined') {
     settings.selections.regions = [];
   }
+  if (typeof settings.selections.weatherWindEnabled === 'undefined') {
+    settings.selections.weatherWindEnabled = false;
+  }
+  if (typeof settings.selections.oceanCurrentsEnabled === 'undefined') {
+    settings.selections.oceanCurrentsEnabled = false;
+  }
+
   // ensure legacy notes selections section is removed
   if (typeof (settings as any).selections.notes) {
     delete (settings as any).selections.notes;
@@ -588,7 +597,9 @@ export function defaultConfig(): IAppConfig {
       aisFilterByShipType: false,
       aisState: [], // list of ais state values used to filter targets
       resourceSets: {}, // additional resources
-      infolayers: null
+      infolayers: null,
+      weatherWindEnabled: false,
+      oceanCurrentsEnabled: false
     }
   };
 }
--- a/src/app/modules/map/fb-map.component.html
+++ b/src/app/modules/map/fb-map.component.html
@@ -809,6 +809,19 @@
     </fb-vessel-trail>
   }
 
+  <!-- weather overlays -->
+  <fb-weather-wind
+    [show]="app.config.selections.weatherWindEnabled"
+    [opacity]="0.55"
+  >
+  </fb-weather-wind>
+
+  <fb-weather-currents
+    [show]="app.config.selections.oceanCurrentsEnabled"
+    [opacity]="0.55"
+  >
+  </fb-weather-currents>
+
   <!-- chart boundaries-->
   @if (app.data.chartBounds.show) {
     <fb-chart-bounds
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -121,6 +121,8 @@ import { Units } from 'ol/control/ScaleLine';
 import { DragBoxEvent } from 'ol/interaction/DragBox';
 import { MapService } from './ol/lib/map.service';
 import { AppIconDef } from '../icons';
+import { LayerWindWeatherComponent } from './ol/lib/resources/layer-wind-weather.component';
+import { LayerCurrentsWeatherComponent } from './ol/lib/resources/layer-currents-weather.component';
 
 interface IResource {
   id: string;
@@ -166,7 +168,9 @@ enum INTERACTION_MODE {
     ResourcePopoverComponent,
     ResourceSetPopoverComponent,
     VesselPopoverComponent,
-    S57PopoverComponent
+    S57PopoverComponent,
+    LayerWindWeatherComponent,
+    LayerCurrentsWeatherComponent
   ],
   templateUrl: './fb-map.component.html',
   styleUrls: ['./fb-map.component.css']
--- a/src/app/modules/map/ol/lib/resources/layer-currents-weather.component.ts
+++ b/src/app/modules/map/ol/lib/resources/layer-currents-weather.component.ts
@@ -1,0 +1,235 @@
+import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  Input,
+  OnChanges,
+  OnDestroy,
+  SimpleChanges
+} from '@angular/core';
+import Feature from 'ol/Feature';
+import Point from 'ol/geom/Point';
+import VectorLayer from 'ol/layer/Vector';
+import VectorSource from 'ol/source/Vector';
+import { Fill, Icon, Stroke, Style, Text } from 'ol/style';
+import {
+  Observable,
+  Subject,
+  Subscription,
+  catchError,
+  debounceTime,
+  of,
+  switchMap,
+  tap
+} from 'rxjs';
+import { FeatureLike } from 'ol/Feature';
+import { fromLonLat, transformExtent } from 'ol/proj';
+
+import {
+  WeatherService,
+  OceanCurrentSample
+} from 'src/app/modules/weather/weather.service';
+import { MapComponent } from '../map.component';
+
+@Component({
+  selector: 'ol-map > fb-weather-currents',
+  template: '<ng-content></ng-content>',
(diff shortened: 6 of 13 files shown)


joelkoz:
Hi @Tominouu — first off, thank you for this. 👏 This is a genuinely useful idea, and it's clear you put real care into it: per-viewport sampling to keep API calls down, opacity controls, clean opt-in toggles, keyless/global-coverage sources. For a first contribution to Freeboard this is a strong piece of work, and I'd very much like to find the right home for it.

Before we go to a line-by-line review, I want to share two pieces of Signal K architecture that aren't obvious from the outside — because I think they make your feature *more* valuable, not less, and may save you some maintenance burden down the road.

### 1. Rain radar → this belongs in a server-side chart provider, and that's a big win for you

Signal K already has an official, proven pattern for serving tile layers like RainViewer: [`@signalk/charts-plugin`](https://github.com/SignalK/charts-plugin). It has an **"Online chart providers"** option where a tile source is registered by config alone — `type: tilelayer` plus a URL containing `{z}`, `{x}`, `{y}` (the schema's own example is literally `http://example.org/{z}/{x}/{y}.png`). That's *exactly* the shape of your RainViewer layer — and notably, this path involves **no MBTiles, no downloading, no conversion**; it's pure tile-serving, so it's the simplest corner of that plugin to model on.

The big payoff: a chart provider is consumed by the charts API, so **the radar would then work in *any* Signal K chart plotter — not just Freeboard** — and the server handles refresh once for every connected client instead of each browser polling RainViewer on its own.

There's one nice wrinkle that actually makes a small dedicated plugin worthwhile rather than a plain config entry: RainViewer's tile path includes a "latest frame" timestamp that rotates every ~10 minutes, so a static URL would go stale. A tiny provider plugin that resolves the current frame and registers the live URL is the real value-add — and the `OnlineChartProvider` registration in charts-plugin ([`src/types.ts`](https://github.com/SignalK/charts-plugin/blob/master/src/types.ts), [`src/index.ts`](https://github.com/SignalK/charts-plugin/blob/master/src/index.ts)) is a compact, working template to start from. If you (or an AI assistant) clone that shape, you've got a publishable `signalk-rainviewer-charts` plugin in not much more than the code you've already written.

### 2. Wind & currents → the *rendering* is the valuable new part; the *data* already has a home

The animated wind/current arrow field is genuinely new and nice — nothing in the ecosystem draws a viewport glyph field like that today, so that rendering is worth keeping in Freeboard.

The one thing to know: Signal K already has a **Weather API**, and its default provider is **Open-Meteo** — the official [`@signalk/open-meteo-provider`](https://www.npmjs.com/package/@signalk/open-meteo-provider) plugin. Freeboard already talks to that API for its forecast popup. So rather than calling `api.open-meteo.com` directly from the browser, the arrow layers can sample their grid against the Signal K Weather API. Two nice consequences:

- It automatically respects whatever weather provider the user has configured (Open-Meteo, Meteoblue, Met.no, AccuWeather…), so your overlay "just works" for more people.
- It keeps Freeboard from being pinned to one vendor's endpoints, which is the part I'd otherwise have to ask you to change before merge.

### Suggested path

Could we split this into focused pieces?

1. **Rain radar → a standalone chart-provider plugin** (lives outside Freeboard; benefits the whole ecosystem).
2. **Wind + currents → a Freeboard PR** that keeps your rendering but pulls data from the Signal K Weather API instead of direct Open-Meteo calls.

That keeps each change single-purpose (our contribution standard here) and, honestly, gets your work in front of *more* users than a Freeboard-only feature would.

None of this is a no — it's "let's land this in the way that gives it the longest legs." If you're up for it, I'm glad to walk you through the chart-provider setup and where the Weather API hooks in. Thanks again for jumping in. 🙏

Tominouu:
RainViewer radar has been extracted into a standalone Signal K plugin: signalk-rainviewer-charts 
(https://github.com/Tominouu/signalk-rainviewer-charts)                                            
                                                                                   
It is published on npm and will automatically appear in the Signal K App Store under "Available"   
(keywords signalk-node-server-plugin, signalk-category-charts). It registers as a charts           
resource provider, usable by any SK client — not just Freeboard.                                   
                                                                                                 
Install: npm install signalk-rainviewer-charts or via the App Store.

joelkoz:
Can you update the title and description of the PR to reflect the current state. I can then re-review based on that.

Tominouu:
@joelkoz i updated the title and description

joelkoz:
@coderabbitai full review

joelkoz:
Thanks for turning this around so thoughtfully, @Tominouu — this is exactly the kind of response that makes reviewing a pleasure. 🙌 The wind layer now flows through the Signal K Weather API (so it works with *any* provider), and pulling RainViewer out into [`signalk-rainviewer-charts`](https://github.com/Tominouu/signalk-rainviewer-charts) — published and App-Store-ready — is a real contribution to the whole ecosystem, not just Freeboard. Nicely done.

A few things to sort before this is merge-ready:

### 1. The chart-rendering fallback deserves its own PR — and it's the opposite of a nit

While wiring up your radar plugin you clearly hit something important: charts served by a **server resource provider** omit the `type` field, and without a fallback Freeboard renders them as *nothing*. Your `@else` branch in `fb-map.component.html` fixes that. That's a genuinely valuable catch — it affects **every** chart resource provider, not just weather.

Here's something you may not know as a first-time contributor here: **our changelog is generated from merged PR titles.** So a fix like this needs to ride in its own PR with a title like `fix(charts): render resource-provider charts that omit a type field` — otherwise it's invisible, buried inside a "weather" PR, and to a future reader it looks like an incidental nit rather than the real bug fix it is. It also follows our "one logical change per PR" rule, and it'll get the focused review it deserves (e.g. the current branch is a broad catch-all that renders *any* unrecognized type as raster — worth a moment's thought on whether that's the right default vs. matching the type-less case explicitly).

Could you pull that change out into a separate PR? Happy to fast-track the review on it.

### 2. Ocean currents — the direct Open-Meteo call is fine

Keeping currents on the Open-Meteo Marine API is reasonable: the SK Weather API doesn't expose currents yet, and Freeboard already ships direct third-party sources (the OpenSeaMap seamark layer works the same way), so you're in good company. The only thing I'd ask is a brief code comment noting it's a deliberate stopgap until the SK weather providers expose currents — that way a future contributor knows to migrate it (like you did for wind) rather than wondering why it's different.

### 3. CodeRabbit's full review

I re-ran a full CodeRabbit pass after your last commit and it flagged a handful of items on the weather files — one it marks **critical**, a couple **major**, and the rest minor, and it considers them all quick wins. Could you work through those and resolve them? Once CR is happy and the chart fallback is split out, we'll be in good shape to land this.

Thanks again — this is shaping up really well. 🙏

joelkoz:
Great progress @Tominouu — the rename, the `weatherApi` gate, and narrowing the chart fallback to the type-less case are all exactly what we were after. 👍

Two things to get this green and ready for a final review:

**1. CI is failing on formatting only.** The red checks are all the "Check formatting" step (Prettier) — no test or logic failures. A quick fix:

```
npm run format
git commit -am "style(weather): apply prettier formatting"
git push
```

That should turn the CI matrix green (you can ignore the slow **armv7 / Cerbo GX** job — it runs under emulation and lags far behind the others; it's not a blocker).

**2. The chart-rendering fallback still needs its own PR.** That's the one item from my last note that's still outstanding — the `@else` branch in `fb-map.component.html` that renders resource-provider charts which omit a `type` field. As mentioned, our changelog is generated from merged PR titles, so that fix deserves to land on its own as something like `fix(charts): render resource-provider charts that omit a type field` rather than being folded into this weather PR. Happy to fast-track the review on it once you open it.

Once CI is green and the chart fix is split out, I'll run a full CodeRabbit pass over the final state so we're reviewing the finished code in one go. Thanks again — nearly there! 🙏

Tominouu:
It's okey ! @joelkoz

joelkoz:
You're basically there, @Tominouu — and nice work splitting the chart fix out into #444, that's already merged. 🎉

There's one last cleanup, and it's **fallout from changes on our side, not anything you did wrong.** While this PR was open we merged two things to `master`:

- **#444** — your extracted chart-rendering fallback, now in `master`.
- **#440** — a housekeeping PR that reformatted a batch of existing files to our canonical Prettier style *and* switched on the CI "Check formatting" gate.

Your branch was created before both, so when compared against current `master` it now shows two bits of noise that shouldn't be in a weather PR:

1. It **re-adds the `@else` chart fallback** that #444 already merged (it'd become a duplicate).
2. It **reverts #440's formatting** on ~7 unrelated files (`icons/app.icons.ts`, `symbol-ref.ts`, `symbols.service.ts`, `radar-api.service.ts`, the two notes dialogs, `waypoint-dialog.ts`) — because your local Prettier formatted them differently than the version #440 standardized on.

One action fixes all of it — **rebase onto the latest `master`:**

```
git fetch upstream
git rebase upstream/master
# resolve the fb-map.component.html conflict by keeping master's @else
# (your weather overlay tags stay; the duplicate fallback goes)
git push --force-with-lease
```

After the rebase, please run the formatter using the project's **own** pinned Prettier so it matches our baseline:

```
npx prettier --ignore-path .gitignore --write "src/**/*.+(ts|html)"
```

(Running a globally-installed Prettier is what caused the formatting drift, so `npx` from the repo is the key detail.)

That should leave #413 as a clean, weather-only change: the two `<fb-weather-*>` overlays, the wind + currents layer components, `weather.service.ts`, the weather panel, and the config/type additions — nothing else.

Once that's pushed and CI is green, I'll run a full CodeRabbit pass over the final commit and we'll get this merged. Thanks for bearing with the moving target on our end. 🙏

Tominouu:
@joelkoz  Done. Rebased onto latest master, ran the project's pinned Prettier (no changes needed), and force-pushed. The diff is now clean — only the 13 weather‑related files. CI should go green; let me know if you want me to tweak anything else.

joelkoz:
The rebase came out clean — the unrelated formatting churn is gone, CI is fully green, and CodeRabbit is happy. 🎯 One last little thing and this is ready.

There's a leftover from the merge conflict in `fb-map.component.html`. `master` already renders type-less charts as raster — that landed with #444, folded into the last condition:

```html
} @else if (
  c[0] === 'openstreetmap' ||
  c[1].type?.toLowerCase() === 'tilelayer' ||
  !c[1]?.type            // ← type-less charts handled here
) {
  <fb-tilelayer-raster ... >
}
```

But this branch still has an extra catch-all `@else` right after it:

```html
} @else {                // ← please remove this block
  <fb-tilelayer-raster ... >
}
```

That block is redundant (the `!c[1]?.type` case above already covers type-less charts), and it also quietly re-widens the behavior to render *any* unrecognized chart type as raster — which is the broad catch-all we intentionally narrowed when the fix moved into #444. Could you delete just that trailing `@else { … }` block?

After that, this file's only change is the two `<fb-weather-*>` overlay tags, and the rest of the PR is clean weather-only code. Once it's pushed I'll do a final pass and we'll merge. Almost there — thanks! 🙏

joelkoz:
@Tominouu - I'm taking a closer look at the title and description of this PR, and I think it needs to be re-worked. It currently reads like a log of what you did to get the PR to the state that it is in, and not "what does this PR add or change in Freeboard" once the dust settles. I think Code Rabbit may have actually written your title: `feat(weather): Added a dedicated Weather panel for toggling wind and ocean currents.`  Does that sound right? The description should then give details on that.  Maybe ask your agent to do a "squash" on your commits, then come up with a good title and description for what remains?

Tominouu:
Done. Removed the redundant @else { ... } block, squashed all commits into a single clean commit. Title and description are now: feat(weather): add dedicated Weather panel for toggling wind and ocean currents. Adds a Weather panel in the resource sidebar with controls to toggle wind and ocean-current overlays on the map. Wind data comes from the Signal K Weather API (/weather/observations/point). Ocean currents uses Open‑Meteo as a stopgap until SK providers expose currents.

joelkoz:
@coderabbitai full review

joelkoz:
This is in great shape now — the rebase, squash, and cleanup all landed cleanly, the code is weather-only, and CI is green. 🙌 Two small things and I think we're ready.

**1. One performance item.** The wind overlay re-samples the viewport on every map move and fires ~20 concurrent `/weather/observations` requests to the Signal K server with no debounce. It's opt-in and off by default, so this only happens when a user enables it — but panning around would hammer their server. Could you add a short debounce (~300–500 ms) on the `moveend` handler so a burst of pans collapses into a single fetch? That's the one substantive item worth acting on.

*(Heads-up on CodeRabbit: most of its inline comments here are stale — several point at `instruments/*` files that aren't part of this PR, and one references the `@else` block you already removed. Don't chase those; the debounce above is the real one.)*

**2. The description needs an end-user rewrite.** This is the last thing, and it matters because **the PR title becomes the App Store changelog entry** and the description is what future maintainers read. Right now it reads like implementation notes. Please follow the repo's PR template (`.github/pull_request_template.md`) and:

- Describe it as release notes for a boater — *"adds wind and ocean-current overlays you can toggle from a new Weather panel"* — using the template's three headings.
- **Drop the file list** — GitHub already shows the diff.
- **Drop RainViewer entirely.** It only existed in an earlier draft of this PR and never shipped in Freeboard, so from a user's perspective there's nothing to "extract" — a changelog line mentioning it wouldn't make sense to anyone who wasn't following this PR.
- Suggested title: `feat(weather): add wind and ocean current overlays`.

If it's easier, feel free to hand this to your AI assistant:

```
Rewrite this Freeboard-SK PR's title and description, following .github/pull_request_template.md.
- Title = a user-facing changelog line: "feat(weather): add wind and ocean current overlays".
- Description = end-user release notes (what a boater gets and why), NOT implementation notes.
- Use the template's three headings: "What problem does this solve?", "How does it work?", "How was this tested?". Keep each short.
- Do NOT list changed files — GitHub shows the diff.
- Do NOT mention RainViewer, Open-Meteo internals, request counts, or unit-conversion formulas.
- Mention the overlays are opt-in and off by default.
```

Once the debounce is in and the description is refreshed, we're ready to merge. Thanks for sticking with this through more rounds than either of us expected — you've handled every one. 🙏

PR #420, opened 2026-06-25
perf: debounce config saves and guard dark-mode updates

dillan:
## Summary

Two small hygiene fixes found during a performance pass (a later, independent
item in the series; not dependent on the other perf PRs).

### Debounce config saves
`saveConfig()` writes the config to localStorage **and** PUTs it to the server
(`signalk.appDataSet`). It was called on **every map pan/zoom release**
(`fb-map.onMapMoveEnd`) and on **every UI-config change** (the `uiConfig`
effect), so actively panning the map produced a burst of localStorage writes and
server requests. Added `saveConfigDebounced()` (1 s) and used it on those two
hot paths.

Measured with `perf-harness/` (Playwright driving a fixed pan/zoom sequence):
**config saves during the gesture dropped from 14 → 2.**

The deliberate, immediate `saveConfig()` calls (explicit setting changes) and the
`movingMap` 30 s save timer are left as-is.

### Guard dark-mode updates
`setDarkTheme()` runs `window.matchMedia(...)` + `classList` writes on **every
realtime SignalK delta**. Track the last-applied state and early-return when the
computed mode is unchanged, skipping the redundant DOM work. (Tiny, but it's on a
per-delta path.)

## Risk
Low. No behavior change for explicit saves; map center/zoom and UI toggles still
persist, just coalesced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dark mode handling so the theme only updates when needed, reducing unnecessary visual updates.
  * Made map and UI settings changes save more smoothly by batching rapid updates into a single save.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -539,23 +539,26 @@ export class AppComponent {
 
   /** ************* */
 
+  private darkThemeApplied: boolean;
   private setDarkTheme() {
     const mq = window.matchMedia('(prefers-color-scheme: dark)');
-
-    if (
+    const enabled =
       (this.app.config.display.darkMode.source === 0 && mq.matches) ||
       (this.app.config.display.darkMode.source === 1 &&
         this.app.data.vessels.self.environment.mode === 'night') ||
-      this.app.config.display.darkMode.source === -1
-    ) {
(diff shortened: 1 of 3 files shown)


joelkoz:
@CodeRabbit full review

PR #437, opened 2026-06-29
fix(ais): orient AIS vessel icons to heading instead of COG

### LINKED ISSUE #415: AIS vessel icons don't line up with reported heading
See attached image. Ferries tend to crab significantly, and it is important their icons show so, to not misinterpret their heading and make undesired corrections. 

Image

zeusec:

AIS boats can be drawn pointing the wrong way. If a target is crabbing, where its heading and its course over ground aren't lined up (think of a ferry getting pushed sideways by current), the icon follows COG instead of heading. So the bow ends up pointing along the track rather than where the boat is actually facing. The IEC 62288 display standard orients the icon by heading for exactly this reason.

Closes #415 

## How does it work?

`parseVesselOther()` was picking COG over heading when it set the icon's `orientation`. I just flipped that to prefer heading, and only fall back to COG when there's no heading at all. Targets that never send heading (plenty of Class B units) look exactly the same as before. The COG vector is still drawn as the dashed line, so now you can see the crab as the angle between the hull and that line.

One note... `parseVesselOther` hardcodes this order and ignores the `preferredPaths.heading` setting, which oddly defaults to COG too. I left it as a plain fix, but if you'd rather have orientation follow that setting, I'm happy to redo it that way.

## How was this tested?

Spun up a local Signal K server, fed it a crabbing target at heading 45° / COG 95°, and pointed Freeboard at it. Before the fix the boat and its popover heading sat at 95°; after, both snap to 45°, while the COG vector keeps pointing east.

### Before
before

### After
after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vessel orientation selection by adjusting fallback precedence, prioritizing heading data (true, then magnetic) before course-over-ground values when other heading fields are missing.
* **Tests**
  * Added coverage for AIS target orientation precedence to ensure correct behavior across combinations of reported heading and course-over-ground data.

### DIFF
--- a/src/app/modules/skstream/skstream.facade.spec.ts
+++ b/src/app/modules/skstream/skstream.facade.spec.ts
@@ -1,0 +1,37 @@
+import { expect, describe, it } from 'vitest';
+import { SKStreamFacade } from './skstream.facade';
+import { SKVessel } from '../skresources/resource-classes';
+
+// AIS icons rotate by `orientation`. A crabbing target's heading and COG differ,
+// so the icon must follow heading, with COG only as a fallback when no heading
+// is reported. Guards the precedence in parseVesselOther against a silent
+// reorder (#415).
+describe('SKStreamFacade.parseVesselOther — AIS target orientation', () => {
+  // Exercise the real (private) method without the heavy DI constructor.
+  const orientationOf = (v: SKVessel): number => {
+    const facade = Object.create(SKStreamFacade.prototype) as unknown as {
+      app: unknown;
+      parseVesselOther: (targets: Map<string, SKVessel>) => void;
+    };
+    facade.app = { useMagnetic: false, data: { vessels: {} } };
+    facade.parseVesselOther(new Map([['ais', v]]));
+    return v.orientation;
+  };
+
+  const vessel = (props: Partial<SKVessel>): SKVessel =>
+    Object.assign(new SKVessel(), props);
+
+  it('prefers true heading when both heading and COG are present', () => {
+    expect(
+      orientationOf(vessel({ headingTrue: Math.PI / 4, cogTrue: 1.658 }))
+    ).toBe(Math.PI / 4);
+  });
+
+  it('falls back to COG when no heading is reported', () => {
+    expect(
+      orientationOf(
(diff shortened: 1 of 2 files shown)


joelkoz:
Thanks for this — really nicely done for a first PR. The title and description are exactly the bar I'd want, the before/after makes the behaviour obvious, and orienting the icon to heading with the COG vector left as the separate dashed line is the correct AIS presentation (it's what IEC 62288 / commercial ECDIS do). The graceful fall-through to COG for Class B targets that don't report heading is the right touch too.

A couple of things before merge:

- **On your `preferredPaths.heading` question** — keep it as the plain fix you've written. FSK's contribution guidelines lean toward "don't add configurability unless it's clearly needed," so I'd rather not wire orientation to that setting in this PR. If someone actually asks for it, it's a clean follow-up.

- **One small test ask.** The repo has a vitest harness (co-located `*.spec.ts`, run with `npm run test:ci`), and precedence logic like this is exactly the kind of thing that can silently regress if the chain ever gets reordered back. A single case would lock it in: a vessel with both `headingTrue` and `cogTrue` set, asserting `orientation === headingTrue`. The seam is `SKStreamFacade.parseVesselOther` in `src/app/modules/skstream/skstream.facade.ts`.

Add that and I think this is good to go.

joelkoz:
Side note: It looks like you may be using a coding agent to help out. FYI, I added some agent based documentation to the repo in PR #434 just last night. If you want to rebase you repo against that before you start, your agent may get some extra prompts (like the desire to have associated test coverage).  I'm interested to know if it helps or hinders...

zeusec:
Added the test just now, and left that setting as is. Your AGENTS.md definitely helps, the only thing that threw me was that it's such a small change and I didn't see earlier PRs with tests for fixes like this, but it makes sense now that the AI guidance was just added.

joelkoz:
> the only thing that threw me was that it's such a small change and I didn't see earlier PRs with tests for fixes like this

No worries! A lot of the earlier PRs didn't have them because no one was doing them.  I only recently updated the CI workflow to match the work being done on the main Signal K server project, so these are all very new "requirements".

Thanks again for the fix!

PR #441, opened 2026-06-30
fix(course): use route.distance for route DTG in kilometer units

### LINKED ISSUE #414: Route following: DTG (route) incorrect.
It seems to reflect DTG (waypoint) 

Image

joelkoz:
When following a route with distance units set to **kilometers**, the route
DTG in the course bar mirrored the next-point (waypoint) DTG instead of the
remaining distance along the whole route.

The kilometer branch of the `route.dtg` calculation divided
`courseCalcs.distance` (distance to next point) rather than
`courseCalcs['route.distance']` (remaining route distance). The
nautical-mile branch was already correct, so this only affected km/m users —
which is why the adjacent `route.ttg` / `route.eta` values were right while
route DTG was wrong (see the report's screenshot: point 7 of 34, route DTG ==
waypoint DTG == 1.1 km).

Adds a co-located unit test covering both the kilometer and nautical-mile
paths; the kilometer assertion fails on the previous code and passes with the
fix.

Closes #414

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected route distance calculations so the displayed remaining route distance now uses the proper source value.
  * Fixed kilometer and nautical mile conversions to ensure route DTG is shown accurately.
  * Added tests covering both metric and nautical-mile scenarios to prevent regressions.

### DIFF
--- a/src/app/modules/course/course.service.spec.ts
+++ b/src/app/modules/course/course.service.spec.ts
@@ -1,0 +1,65 @@
+import { TestBed } from '@angular/core/testing';
+import { signal } from '@angular/core';
+import { beforeEach, describe, it, expect } from 'vitest';
+
+import { CourseService } from './course.service';
+import { SignalKClient } from 'signalk-client-angular';
+import { AppFacade } from 'src/app/app.facade';
+import { SKResourceService, SKVessel } from '../skresources';
+import { Convert } from 'src/app/lib/convert';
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai review

joelkoz:
@coderabbitai ignore

PR #444, opened 2026-06-30
fix(charts): render resource-provider charts that omit a type field

Tominouu:
Chart resource providers serve tiles without a `type` property in the resource JSON.  Without a fallback branch Freeboard renders those charts as nothing — they are invisible on the map.                                                                                                                                       
                                                                                                                                                                                              
This change adds an explicit `@else if (!c[1]?.type)` branch that renders type-less charts via `fb-tilelayer-raster`, restoring                                                                                                                         visibility for any provider-based chart layer.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map rendering so raster tile layers now display correctly in cases where chart type information is missing.
  * Reduced the chance of blank or incomplete map content by adding a fallback rendering path.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/app/modules/map/fb-map.component.html
+++ b/src/app/modules/map/fb-map.component.html
@@ -177,16 +177,10 @@
       >
       </fb-tilelayer-vector>
     } @else if (
-      c[0] === 'openstreetmap' || c[1].type?.toLowerCase() === 'tilelayer'
+      c[0] === 'openstreetmap' ||
+      c[1].type?.toLowerCase() === 'tilelayer' ||
+      !c[1]?.type
     ) {
-      <fb-tilelayer-raster
-        [zIndex]="10 + idx"
-        [chart]="c"
-        [overZoomTiles]="overZoomTiles"
-        [mapMaxZoom]="app.MAP_ZOOM_EXTENT.max"
-      >
-      </fb-tilelayer-raster>
-    } @else {
       <fb-tilelayer-raster
         [zIndex]="10 + idx"
         [chart]="c"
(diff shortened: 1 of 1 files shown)


joelkoz:
@coderabbitai review

joelkoz:
Thanks for this — nice catch, and a well-scoped first PR. A type-less chart resource rendering as nothing is a real problem, and raster is the right fallback: historically plenty of chart providers served plain z/x/y tiles without stamping a `type`, so this restores them cleanly. The title and diff are exactly the shape I like to see.

One change before I merge: the new branch is identical to the existing raster branch just above it (lines 182–188) — same four bindings, only the guard differs. Rather than carry two copies of the same `fb-tilelayer-raster` block, fold the new condition into that branch's guard:

```
} @else if (
  c[0] === 'openstreetmap' ||
  c[1].type?.toLowerCase() === 'tilelayer' ||
  !c[1]?.type
) {
```

That's functionally identical (the bodies are the same) and keeps the template from duplicating the renderer. It's the same consolidation CodeRabbit suggested.

One question so I'm sure the fallback is aimed right: which chart provider surfaced this for you? Just want to confirm the type-less charts are genuinely raster tile sources and not, say, a vector/tilejson source that's separately mis-tagged.

Squash that down and I'll get it merged.

Tominouu:
@joelkoz  Done and pushed. I folded the !c[1]?.type check into the existing else if guard so there's no duplicate fb-tilelayer-raster block.                 
                                      
The type‑less charts come from a server‑side ResourceProvider (type: 'charts') that returns JSON without a type field on each chart entry. The tiles are plain z/x/y raster sources — not vector or TileJSON — so the raster fallback is the right target.

PR #445, opened 2026-07-01
docs(signalk): document unit preferences for value display

joelkoz:
## Why

Freeboard converts units client-side via the `Convert` class (`src/app/lib/convert.ts`), independent of the Signal K server's per-user **Unit Preferences**. New UI that displays a value should honor the user's server-side preference instead, so it follows their chosen units (and any later change) with no app-side settings.

## What

Agent/developer documentation only — no code change:

- Add `docs/signalk/unit-preferences.md` — the `displayUnits` metadata contract, unit categories/base units, REST + `sendMeta:'all'` delivery, and the consume-and-convert pattern.
- `AGENTS.md` — a Code-quality guardrail and a Documentation-index entry pointing to the new doc.

@coderabbitai ignore

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,11 @@ for running a local Signal K server to develop against.
 - **Logging.** Use `console.warn` (not `console.error`) for *recoverable*
   feature-detection failures, and `this.app.debug()` for internal state tracing —
   not `console.log`.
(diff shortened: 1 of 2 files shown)


joelkoz:
Wrong solution. Extending `formatValueForDisplay()` instead

Issue #304, opened 2025-11-20
Separate unit for wind speed


tde01065:
In some locales, wind speed and boat speed are communicated in different units. In Sweden, we tend to use knots for boat speed and m/s for wind speed. Currently, it's only possible to set one unit for all speed measurements.

<img width="1571" height="124" alt="Image" src="https://github.com/user-attachments/assets/993c7f30-02d7-4a08-8d9d-f9872b007658" />

panaaj:
v2.20.0 has initial support for Server Unit preferences (signalk-server v2.23.0).
Future releases will have additional support for per path unit setting.

joelkoz:
@tkurki - I see you referenced a "preferred units" feature on sk server and closed it as un-merged?  Was the feature rejected? Was it superseded by something different?  I can add "per path" display units to FSK, but having the configuration server side is better (and more universal).  Where does this stand?

tkurki:
Adrian's last comment - see 
- https://demo.signalk.io/admin/#/documentation/Guides/Unit_Preferences.html
- https://demo.signalk.io/admin/#/documentation/Developing/Unit_Preferences.html

PR #446, opened 2026-07-02
feat(units): support per-path display units in formatValueForDisplay

joelkoz:
## Why

Freeboard applies the server's unit preferences at the **category** level only (all speeds, all depths, …). It can't yet honor a **per-path** override — e.g. wind speed in m/s while boat speed is in knots (#304) — which the server expresses via `meta.displayUnits`.

## What

Extend the existing display funnel `AppFacade.formatValueForDisplay()` with an optional `path`:

- When `path` is supplied **and** a per-path `displayUnits` override is cached for it, that override wins; otherwise the current category preset applies.
- `path` is optional, so all existing callers and derived (path-less) values are unaffected — no behavior change until per-path overrides are populated.
- Adds the `SKPathDisplayUnits` type, a `pathDisplayUnits` cache with `get/setPathDisplayUnits()` accessors, and unit tests.

Also adds contributor guidance (`AGENTS.md` + `docs/signalk/unit-preferences.md`): display user-facing values through the funnel, not `Convert` directly (`Convert` stays the pure primitive for geometry/numeric math).

This is the foundation; wiring the cache to live path metadata (`sendMeta: 'all'`) and the #304 fix follow separately.

## Summary by CodeRabbit

* **New Features**
  * Added support for per-path display unit preferences, so numeric values can now follow more specific unit settings when shown in the UI.
  * Improved value formatting so user-facing numbers can respect path-specific overrides, symbol choices, and formatting options.

* **Documentation**
  * Expanded guidance on how numeric values are displayed and how unit preferences are resolved.

* **Tests**
  * Added coverage for formatting behavior across default, override, and fallback scenarios.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,14 @@ for running a local Signal K server to develop against.
 - **Logging.** Use `console.warn` (not `console.error`) for *recoverable*
   feature-detection failures, and `this.app.debug()` for internal state tracing —
   not `console.log`.
+- **Displaying values → `formatValueForDisplay()`.** Render every user-facing numeric
+  value through `app.formatValueForDisplay(value, sourceUnit, { path })`. It applies the
+  user's unit preferences — a per-path server override (`meta.displayUnits`) when a
+  `path` is given, otherwise the category preset — and returns the value with its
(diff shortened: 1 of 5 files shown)


joelkoz:
@coderabbitai ignore

PR #448, opened 2026-07-02
feat(map): add per-vessel AIS track display

### LINKED ISSUE #374: Individual vessel track display
Add the ability to select a vessel, and have it display its track. This could include its old track if that is a possibility.

joelkoz:
Adds the ability to display an **individual** AIS vessel's track, rather than only
the all-or-nothing "show tracks for all vessels" setting.

Users reported that showing every vessel's track is too cluttered when they only
want to follow one boat. This lets them pick specific vessels.

**How it works**
- A session-only selection set (`app.data.vessels.showTrack`) drives display,
  mirroring the existing `flagged` pattern — nothing is persisted to config.
- With the global "show all tracks" setting **on**, all tracks render as before.
  With it **off**, only the selected vessels' tracks render.
- Toggle a vessel's track from either the **vessel popover** (on the map) or the
  **vessels list**, sharing the same selection.
- The track layer mounts whenever the global toggle is on *or* at least one vessel
  is individually selected.

Track data continues to come from the server tracks plugin (already fetched
per-vessel), so no server-side change is required.

Selection logic is extracted to a small pure helper (`src/app/lib/vessel-track.ts`)
with unit tests.

Closes #374

## Summary by CodeRabbit

* **New Features**
  * Added a vessel track visibility toggle in vessel details and AIS lists.
  * You can now show selected vessel tracks for the current session, with support for viewing all tracks or only chosen ones.
  * Map overlays now respect track visibility settings and update dynamically when selections change.

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -627,7 +627,8 @@ export function initData(): FBAppData {
       active: null,
       closest: [],
       prefAvailablePaths: {}, // preference paths available from source,
-      flagged: [] // flagged ais targets
+      flagged: [], // flagged ais targets
+      showTrack: [] // ais targets to display track for (session-only)
     },
     aircraft: new Map(), // received AIS aircraft data
     atons: new Map(), // received AIS AtoN data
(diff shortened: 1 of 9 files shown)


joelkoz:
@coderabbitai review

PR #491, opened 2026-07-08
feat(weather): add tidal currents overlay with time-stepping controls

marcelrv:
Display tidal current arrow field on the map using `signalk-tidal-currents` plugin data. Includes:

- TidalCurrentsService: state (scrubTime, play/pause), 24h bar tick/label math, API via SignalKClient
- TidalCurrentsLayerComponent: OpenLayers vector layer with speed-colored arrows, auditTime(300) pipe
- Weather panel: clickable/draggable 24h time bar, play/pause/step/reset buttons, pill label
- Feature flag (tidalApi) gated on plugin detection via /signalk/v2/features
- Dark/light theme support via app.uiConfig().invertColor
- Click handling delegated to Freeboard's processMapClick for consistent popovers

## Screenshot
tidal currents overlay

## Summary by CodeRabbit

* **New Features**
  * Added tidal currents support across the weather and map experience, including a 24h time scrub bar with playback controls and a new tidal point layer on the map.
  * Added a settings toggle for tidal currents, gated by a SignalK plugin/feature-flag detection.
  * Tidal map clicks now show a compact popover with formatted speed and direction.

* **Bug Fixes**
  * Backfilled missing tidal-currents selection settings to default to off.
  * Improved timeline “now” behavior when not actively scrubbing.

* **Style**
  * Added a compact popover variant for tighter tidal point display.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -832,7 +832,8 @@ export class AppComponent {
       weatherApi: false,
       radarApi: false,
       notificationApi: false,
-      buddyList: false
+      buddyList: false,
+      tidalApi: false
     };
     this.signalk.get('/signalk/v2/features?enabled=1').subscribe(
       (res: {
@@ -866,6 +867,11 @@ export class AppComponent {
           if (p.id === 'signalk-pmtiles-plugin') {
             this.app.debug('*** found PMTiles plugin');
             hasPlugin.pmTiles = true;
+          }
+          // tidal currents
+          if (p.id === 'signalk-tidal-currents') {
+            this.app.debug('*** found signalk-tidal-currents plugin');
+            ff.tidalApi = true;
           }
         });
         this.app.featureFlags.update((current) => {
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -371,7 +371,8 @@ export function cleanConfig(
       resourceSets: {},
       infolayers: null,
       weatherWindEnabled: false,
-      oceanCurrentsEnabled: false
+      oceanCurrentsEnabled: false,
+      tidalCurrentsEnabled: false
     };
   }
 
@@ -412,6 +413,9 @@ export function cleanConfig(
   }
   if (typeof settings.selections.oceanCurrentsEnabled === 'undefined') {
     settings.selections.oceanCurrentsEnabled = false;
+  }
+  if (typeof settings.selections.tidalCurrentsEnabled === 'undefined') {
+    settings.selections.tidalCurrentsEnabled = false;
   }
 
   // ensure legacy notes selections section is removed
@@ -618,7 +622,8 @@ export function defaultConfig(): IAppConfig {
       resourceSets: {}, // additional resources
       infolayers: null,
       weatherWindEnabled: false,
-      oceanCurrentsEnabled: false
+      oceanCurrentsEnabled: false,
+      tidalCurrentsEnabled: false
     }
   };
 }
--- a/src/app/app.facade.ts
+++ b/src/app/app.facade.ts
@@ -226,6 +226,7 @@ export class AppFacade extends InfoService {
     resourceTracks: boolean;
     infoLayers: boolean;
     buddyList: boolean;
+    tidalApi: boolean;
   }>({
     anchorApi: true, // default true until API is available
     autopilotApi: false,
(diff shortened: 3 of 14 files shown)


joelkoz:
Thanks for this, Marcel — this is a genuinely useful addition. A time-scrubbable
tidal-current field is exactly the kind of passage-planning tool the coastal-cruising
crowd (especially high-tidal-range waters) will get real value from, and gating it on
plugin detection so it stays off-by-default for everyone else is the right call. Nicely
scoped as a single logical change, too.

A few things before it's ready to merge:

**CI — it's just formatting.** All the platform jobs are failing on the one
`Check formatting` (prettier) step; everything substantive (build, lint, plugin
lifecycle, npm-pack) passes, and the tests are only skipped because formatting fails
first. Running the repo's pinned prettier over the changed files and pushing should turn
the whole matrix green. (Heads-up on the armv7 / Cerbo GX job — it runs under QEMU and is
always much slower, so don't wait on it once the rest are green.)

**Unit preferences.** The speed/direction values in the arrow labels and popover go
through `toFixed` with an inline knots→m/s constant. FSK's convention is to funnel every
user-facing number through `app.formatValueForDisplay()` so it honors the user's display-
unit settings (e.g. a user set to m/s or km/h should see that automatically). Worth
routing the tidal values through that same path — CodeRabbit flagged the duplicated
conversion in a couple of spots too.

**Tidal map-click popover.** CodeRabbit noted that `formatPopover` has no `case 'tidal'`,
so clicks routed there may fall through without rendering. Worth a quick check that
clicking an arrow actually shows the popover end-to-end.

**One note on CodeRabbit:** your last commit (the `dragEnd$` cleanup) hit CR's review
rate limit, so it never actually got reviewed even though the check shows green. Your
next push (the formatting fix) will re-trigger CR and it'll pick that commit up then.

**A screenshot or short clip** of the overlay + scrub bar in action would help a lot —
both for review and for anyone browsing the PR later.

And a small future-proofing thought: since `signalk-tidal-currents` is still early (0.x),
it'd be good to note which grid-API shape this targets so a later plugin bump doesn't
silently break the overlay.

Really nice work — looking forward to getting this in.

joelkoz:
Two follow-ups on the CodeRabbit threads, now that I've gone through their
resolved/unresolved state:

- **Still open:** the step-button tooltips in `weatherlist.html` read `-6h`/`+6h`,
  but `stepBackBig()` / `stepForwardBig()` only move by 1h — so either the tooltips
  or the step size should change to match.
- On my earlier unit-preferences note — looks like you've already routed the
  layer-side values through `formatValueForDisplay()` (that thread's resolved), so
  disregard that part. The only thing worth a glance is whether the click-popover
  path in `fb-map.component.ts` still has its own inline knots→m/s conversion rather
  than reusing the same helper.

So the two genuinely-open functional items are that tooltip mismatch and the missing
`case 'tidal'` in `formatPopover`.

marcelrv:
Yes, the popup is working. I included it in the screenshot.
I think the 6h jump / mouseover was fixed before. the small step however was not in syc, the real increment was 15min and the mouseover said 1h. Now the real jump is synced to 1h.
 

image

marcelrv:
And screenshot with the reason for the tidal current side project...

tide aware route planning... providing the best route and best departure time using the tide current... 
I always hate doing the calculations how the tide affect the route (what is the best route considering the current) and even the commercial apps I used did not support there...

image

joelkoz:
> And screenshot with the reason for the tidal current side project...
> 
> tide aware route planning... providing the best route and best departure time using the tide current... I always hate doing the calculations how the tide affect the route (what is the best route considering the current) and even the commercial apps I used did not support there...
 
 Nice! I assume that was done with the "Plotter Extensions" API?  That is a great use of it!
 
 One hint: you may want to name your extension `Autoroute - tidal currents`. I saw another user working on a different auto router, and I have a cheap reference once, and they all share the same name.  I will probably have to come up with some mechanism to distinguish plugins that use the same window titles.

joelkoz:
fyi - I just cut 3.0.0-beta.3, and this made it in.

marcelrv:
> > And screenshot with the reason for the tidal current side project...
> > tide aware route planning... providing the best route and best departure time using the tide current... I always hate doing the calculations how the tide affect the route (what is the best route considering the current) and even the commercial apps I used did not support there...
> 
> Nice! I assume that was done with the "Plotter Extensions" API? That is a great use of it!
> 
> One hint: you may want to name your extension `Autoroute - tidal currents`. I saw another user working on a different auto router, and I have a cheap reference once, and they all share the same name. I will probably have to come up with some mechanism to distinguish plugins that use the same window titles.

Yes, i have some ideas for the naming.. the autoroute name was from before the sudden explosion of autoroute plugins 😊

Issue #457, opened 2026-07-03
[FR] exposure and contrast controls for satellite/aerial image layers


tkurki:
when scouting anchorages  and places where you can tie up with a stern anchor it would be useful to be able to adjust the contrast and exposure of the image layers. this way you can fish out all the possible details from images where land may have ok tones but water is less than optimal

<img width="1496" height="735" alt="Image" src="https://github.com/user-attachments/assets/7d02880f-7d51-44c5-a9c2-0ba4ba4e97f3" />

joelkoz:
Thanks @tkurki  — nice idea, and that before/after shot makes the case on its own. I'm going to take an initial pass at this, but wanted to check the approach with you first.

**Scope.** My instinct is to start with the 80% win: **brightness + contrast** (plus saturation) applied as filters over the existing raster layer. That's a small, low-risk change that works on every raster/aerial layer, and in practice "exposure" is mostly covered by brightness. Is that **enough** for what you're after?

The fuller set — true `exposure` + `gamma` on top — is available, but only via OpenLayers' WebGLTile styling, which would mean migrating the XYZ raster path from the canvas tile layer to WebGL rendering. That's a noticeably bigger blast radius (it changes how all raster charts render, not just the adjustment knobs), so I'd rather not take it on unless you think those extra controls genuinely earn it. If you do, say the word and I'll have a go — just flagging it's more risk for the marginal gain.

**Persistence.** I can do this either way; the real question is what's more convenient for you as the user. Two options:
- **Live/session only** — a refresh resets everything back to default.
- **Persisted per-chart** — your adjustments stick across reloads (handy for an anchorage you revisit), with a **Reset** button to snap back to defaults.

Which feels better in practice? I'm leaning toward persist + reset, but you're closer to the use case.

Once you weigh in I'll get a first cut up.

tkurki:
In my experience these are very tile-dependent - different images have different sun angles, have been captured during different times in the year and water qualities differ wildly. So I think the best option is that the filters are applied **only when you open a UI element that has the controls* for them* (sliders would work great) and the values are persisted per layer, so that an accidental reload does not loose what you were looking at completely.

Brightness & contrast seem the best option to me, anything beyond that is too complex for the value.

joelkoz:
@tkurki - #489 is my first stab at implementation. I've released `3.0.0-beta.2` to the app store and it is part of that. Give it a try. If there are any problems, re-open this issue.

tkurki:
[...]

PR #497, opened 2026-07-10
feat(charts): make image adjustment a modeless draggable palette

joelkoz:
Reworks the Image Adjustment control based on feedback on #457.

Previously it was a centred, modal dialog that covered the area of interest and blocked the map. Now:

- **Modeless & draggable** — a compact palette anchored top-left with no backdrop, so panning and zooming stay available while adjusting. Drag it by the title bar to reposition.
- **Opens with the chart list closed** — clicking the adjust control closes the chart list to free the map. Ownership moved to `SKResourceService` so the palette outlives the chart list.
- **Compact** — smaller footprint; keeps a title + chart-layer name for context.
- Sliders stay live; **SAVE** persists per-chart, cancel reverts, **RESET** returns to neutral.

## Screenshot
Modeless image adjustment palette

Addresses the UX points raised on #457. Tests cover the relocated save/cancel persistence (including the config-vs-chart fallback).

### DIFF
--- a/src/app/lib/components/dialogs/image-adjustment-dialog.ts
+++ b/src/app/lib/components/dialogs/image-adjustment-dialog.ts
@@ -8,6 +8,7 @@ import {
   MatDialogRef,
   MAT_DIALOG_DATA
 } from '@angular/material/dialog';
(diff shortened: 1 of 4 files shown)


joelkoz:
@coderabbitai pause

PR #508, opened 2026-07-12
feat(plotterext): support reverse embedding (embedding host)

### LINKED ISSUE #507: feat(plotterext): implement reverse embedding (embedding host) in the reference host
Implement the **reverse-embedding / embedding-host** mode in the reference host (Freeboard-SK), so an application embedding FSK in an iframe (e.g. KIP) can drive it over the Plotter Extensions bus as a caller, instead of the bespoke night-mode-only `postMessage` bridge.

Contract is already published:
- **Spec:** [`docs/api/plotter-extensions-api.md` → *Embedding Hosts*](https://github.com/SignalK/freeboard-sk/blob/master/docs/api/plotter-extensions-api.md) (merged in #506).
- **Bus:** `signalk-plotterext-bus@^0.10.0` — adds the `embedding-host` context kind, optional `bus.ready` `{ id }`, `HostConnection.adoptCallerId`, and `connectExtension.id`.

Design decisions (from #503): FSK stays the API host/provider; the embedding app is the caller and initiates; **full host API surface** exposed; **same-origin origin pin is the only trust gate** (never `*`); caller identity via **Option B** — the embedder asserts its own id in `bus.ready` and FSK adopts it (`adoptCallerId: true`), falling back to the synthetic `embedding-host` only when the caller sends none.

### Where it lands (recon)

- **`src/app/app.facade.ts`** — embed detection already exists: `isTopWindow()` and the `if (!this.isTopWindow())` branch in the constructor that currently wires the old `parseMessageFromParent` bridge. Stand up the embedding-host connection here. **Keep the old bridge as-is** (backward compat — it and the shipped `nightMode` capability both remain until KIP migrates).
- **`src/app/modules/plotterext/plotterext.service.ts`** — `attachBackground()` is the template. Add a sibling, e.g. `attachEmbeddingHost()`:
  ```ts
  new HostConnection({
    port: windowPort(window.parent, { origin: <FSK's own served origin> }), // NOT '*'
    hostInfo: this.hostInfo(),                 // full HOST_CAPABILITIES
    context: { kind: 'embedding-host', id: 'embedding-host', instanceId: null },
    adoptCallerId: true,
    methods: { ...state, ...signalk, ...units, ...resources, ...map, ...route, ...chart, ...uiPanel } // same set attachBackground registers
  })
  ```
- **`package.json`** — bump the bus dep to `^0.10.0`; regenerate the lockfile **from the registry** (`rm -rf node_modules package-lock.json && npm install`) — do not `npm link`/`file:` it (poisons `npm ci` in CI).

### Two implementation nuances to settle

1. **Origin pin source.** `windowPort` defaults `origin` to the listening window's own origin, which is the same-origin gate we want — just don't pass `'*'`. Reconcile against the old bridge's `event.origin === this.hostDef.url` check and pick the correct pin (FSK's served origin vs. the configured server URL); they coincide for the same-server embed but decide deliberately.
2. **State namespacing.** `attachBackground` binds `stateMethods(extension, null)` at construction, but the caller id only arrives in `bus.ready`. For the MVP, namespacing embedding-host `state.*` under a fixed `embedding-host` key is acceptable (the adopted id is then used for handshake `context.id` / attribution). Lazily namespacing by the adopted id is a nice-to-have, not required.

### Not in scope

- **`fsk-mcp` sync — N/A.** This mode adds **no new host API methods** (it exposes the existing surface to a parent), so there is nothing new to mirror in `dev-tools/fsk-mcp`.
- The KIP side (consuming `connectExtension({ port: windowPort(fskIframe.contentWindow, { origin }) })`) lives in KIP's repo and is coordinated separately.

Closes the FSK half of #503.

joelkoz:
When Freeboard runs inside another application's iframe (an "embedding host", e.g. KIP), expose the full host API to that parent over the Plotter Extensions bus — the reverse of the usual topology. Freeboard stays the API host but points its port at `window.parent`; the embedding host is the caller and initiates the handshake. This lets an embedder drive Freeboard through the full negotiated surface instead of the bespoke night-mode-only `postMessage` bridge.

- `attachEmbeddingHost()` stands up one `HostConnection` targeting `window.parent`, origin-pinned to Freeboard's own origin (same-origin embedders only — they already share the user's session), with `adoptCallerId: true` so the caller's asserted id becomes the handshake `context.id`. `init()` calls it once, only when embedded (`!isTopWindow()`).
- The legacy `parseMessageFromParent` night-mode bridge is retained for backward compatibility.

Implements the reference-host side of the API spec's *Embedding Hosts* section (added in #506); requires `signalk-plotterext-bus@^0.10.0`.

Closes #507.

## Summary by CodeRabbit

* **New Features**
  * Freeboard can now run inside another application’s iframe while exposing its Plotter Extensions host API to the embedding application.
  * Added support for embedding-host contexts, caller identity adoption, state sharing, and the full host capability set.
  * Same-origin access is enforced for embedded connections.

* **Documentation**
  * Added guidance for configuring and using reverse embedding, including state scoping and legacy night-mode compatibility.

* **Tests**
  * Added coverage for embedded connections, host capabilities, caller IDs, and state synchronization.

### DIFF
--- a/docs/freeboard/freeboard-plotter-ext-support.md
+++ b/docs/freeboard/freeboard-plotter-ext-support.md
@@ -43,6 +43,27 @@ package; Freeboard depends on it (`^0.8.0`) and imports its host entry point
 (The authoritative list is `HOST_CAPABILITIES` in
 `src/app/modules/plotterext/types.ts`.)
 
+### Embedding hosts (reverse embedding)
+
+Freeboard can also run **inside** another application's iframe (an *embedding
+host*, e.g. KIP) and expose the same host API to that parent — the reverse of the
+usual topology. See the API spec's *Embedding Hosts* section for the contract.
+
+- `PlotterExtensionService.attachEmbeddingHost()` registers one `HostConnection`
+  whose `windowPort` targets `window.parent`, with the full method set and a
+  `kind: 'embedding-host'` context. `init()` calls it once, only when embedded
+  (`!app.isTopWindow()`). Freeboard only stands up the endpoint — the **embedding
+  application initiates** the handshake by calling `connectExtension` from the
+  parent frame.
+- **Same-origin only.** The port pins the origin to Freeboard's own
+  (`window.location.origin`) — a cross-origin embedder is refused. A same-origin
+  embedder already shares the user's session, so this grants no new authority.
+- **Caller identity.** `adoptCallerId: true` adopts the id the embedder asserts in
+  `bus.ready` as the handshake `context.id`; `state.*` stays under a single
+  `embedding-host` scope (one embedder per server in practice).
+- The legacy parent-`postMessage` night-mode bridge in `app.facade.ts`
+  (`parseMessageFromParent`) is retained for backward compatibility.
+
 ## The `routes` capability
 
 `routes` lets an extension create and edit routes on the chart — e.g. an
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@signalk/server-api": "^2.24.0",
         "geolib": "^3.3.3",
         "google-protobuf": "^4.0.1",
-        "signalk-plotterext-bus": "^0.9.0",
+        "signalk-plotterext-bus": "^0.10.0",
         "tslib": "^2.0.0",
         "uuid": "^14.0.1"
       },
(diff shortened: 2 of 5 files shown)


joelkoz:
@coderabbitai pause

PR #510, opened 2026-07-12
feat(charts): trim image adjustment palette and remember its position

joelkoz:
Follow-up refinements to the modeless Image Adjustment palette (#497), addressing further feedback on #457.

- **Header trimmed** — drops the redundant "Image Adjustment" title and shows just the chart-layer name.
- **Tighter layout** — reduced whitespace between the slider rows and around the action bar; RESET/SAVE are flat text buttons.
- **Remembers its position** — the palette reopens where it was last dragged; the drag offset is persisted in app config so it survives reloads and sessions.

Tests cover the position persistence (opens at the remembered offset; persists on drag).

## Before / after

| Before | After |
|---|---|
| before | after |

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -44,6 +44,9 @@ export function cleanConfig(
       settings.ui.showNotes = true;
     }
   }
+  if (typeof settings.imageAdjustPalettePos === 'undefined') {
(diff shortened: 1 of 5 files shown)


joelkoz:
@coderabbitai pause

PR #576, opened 2026-07-25
fix(map): keep popups and route editing aligned when the map is scrolled a full world sideways

### LINKED ISSUE #572: Popups and route-vertex editing target the wrong world copy
### What happens

Pan the map one world east or west (there's no limit on horizontal panning), then click any feature — a route waypoint, a vessel, an AtoN. The popup opens, but in a *different copy of the world*: pan back and you'll find it there. It affects every feature type, at every location. It has nothing to do with the antimeridian — a route in the Florida Keys reproduces it.

The same failure hits **route editing**, not just popups. In Modify mode you can only grab and drag a route's vertices while viewing the primary world — pan one copy over and the vertices are dead to the pointer. Fixing the popups alone would leave this untouched, because it's the same root cause on the *hit-test* side rather than the placement side.

### Root cause

Click coordinates are normalised on the way in, then rebuilt as a map position on the way out:

```
click at 197°E  →  internal x ≈ 22,015,000
map.component.ts:390    lonlat: toLonLat(event.coordinate)   →  −163°   (world offset discarded)
overlay.component.ts:49 setPosition(fromLonLat(position))    →  x ≈ −18,124,000
```

`toLonLat` deliberately wraps anything outside ±180 back into range, so which copy the user was looking at is lost by the time the overlay is placed. `fromLonLat` then rebuilds it in the primary world — exactly one world width (40,075,016 m) away.

Placement isn't the only casualty. OL's `Modify` interaction indexes the feature's **canonical** geometry (`interaction/Modify.js` → `feature.getGeometry()`) and hit-tests the pointer against it. A click in another world copy lands in that copy's Mercator space, which doesn't line up with the primary-world geometry — so no vertex matches and the drag does nothing. Same missing concept ("which world copy is the user interacting with"), on the hit-test side instead of placement.

This can't be avoided by restricting the view. OpenLayers' `createCenterConstraint` sets the centre extent to ±Infinity in X whenever the projection is global, regardless of `multiWorld` — infinite horizontal panning is baseline behaviour, and `multiWorld` only governs how far you can zoom out.

### Why this is worth fixing at the source

The underlying gap is that the map code has no explicit distinction between two different coordinate spaces:

- **Render space** — unwrapped, world-copy-aware, OL-internal. Purely a drawing concern.
- **Data space** — canonical WGS84 in `[-180, 180]`. Everything persisted, streamed, or read by another app.

`toLonLat`/`fromLonLat` convert between them while silently *normalising*, which is correct for data and destroys exactly what placement needs. Because that distinction was never made explicit anywhere, the same underlying problem has surfaced in several places over the years and been handled locally each time (see *Existing one-off patches* below).

What made the common cause visible was a fairly deep read through the OpenLayers renderer internals, done with AI assistance. That's a much cheaper investigation today than it was when most of this code was hand-written, and it's the only reason the pattern is apparent now rather than then. A shared notion of "which world am I placing in" would replace the scattered patches.

### Constraint on any fix

**Nothing written to the Signal K server may contain an out-of-range coordinate.** Wrapping is an FSK/OpenLayers rendering construct; stored resources are GeoJSON, and RFC 7946 §3.1.9 expects antimeridian-crossing geometry to be cut rather than expressed with longitudes beyond ±180. A route containing `197` is malformed, and other consumers of the Signal K API are entitled to reject it.

So normalisation must stay at every data boundary. The world offset belongs only to interaction placement/hit-testing — ideally applied in Mercator, so a wrapped value never exists in degrees at all and can't leak into storage.

Three cases need covering:

- popups anchored to the **click point** (S57, bearing/distance, tidal, feature list) — need the click's world offset preserved
- popups anchored to a **stored position** (vessel, AIS, AtoN, SAR, meteo, aircraft) — the canonical position must be translated into the clicked copy, and stay there as the vessel moves
- **route-edit vertex hit-testing** (OL `Modify`) — the interaction indexes the feature's canonical geometry, so a click in another world copy matches no vertex; the clicked point (or the edit geometry) must be shifted into the copy the user is working in. @daniel-freiermuth's `bestShift` search in #364 is a working fix for this side.

### Affected surfaces (implementation map)

Hit-testing splits into two mechanisms that share no code, plus one standalone. Line numbers are as of master `27ad2c41` and will drift — anchor on the function names.

**Do NOT touch (already correct):**

- **Feature picking** — `map.getFeaturesAtPixel(...)` (`map.component.ts` ~365, ~387) is pixel-based and already world-correct: OL renders every visible copy, so the click finds the right feature. The popup *appears* — only its position is wrong. The bug is the coordinate that rides alongside, never the pick.
- **Map-centre emission** — `augmentMoveEvent` → `getMapCenter()` (`map.component.ts` ~409 / ~468) normalises deliberately; a canonical centre is correct for config and display. Leave it.

**Mechanism 1 — pointer → coordinate emission (popups, context menu, cursor).** Four near-identical sites emit `toLonLat(<pointer coord>)`, each discarding the world offset. They are *not* funnelled today (three `augment*` helpers plus one inline) — consolidating them is the single highest-leverage fix:

| Site (`map.component.ts`) | Feeds |
|---|---|
| `augmentClickEvent` (~390) | mapClick / mapSingleClick / mapDblClick |
| `emitRightClickEvent` (~371) | mapRightClick / context menu |
| `augmentPointerEvent` (~433) | mapPointerMove / mapPointerDrag |
| pointerDown handler (~339) | mapPointerDown |

*Fix:* one helper that carries the world offset **alongside** the normalised `lonlat` (e.g. emit `lonlat` in `[-180,180]` for data/display **and** the raw world-aware Mercator coordinate, or an integer world index, for placement). Consumers that store or display keep using `lonlat`; consumers that place an overlay use the offset.

*Output choke (single function):* `overlay.component.ts` `setPosition(fromLonLat(position))` (~49, ~63) — the one place a position becomes a pixel. Apply the world offset **here, in Mercator**, rather than denormalising the lon/lat. Stored-position popups additionally need the canonical position shifted into the clicked copy before reaching this point (case 2 above; @daniel-freiermuth's `wrapPos` in #364).

**Mechanism 2 — OL `Modify` interaction (route vertex editing).** `interaction-modify.component.ts` → OL `Modify` builds its own segment index from `feature.getGeometry()` (canonical, primary world) and hit-tests internally; it never calls `getFeaturesAtPixel`, so it cannot share Mechanism 1's fix. *Fix* at the Modify setup (`fb-map.component.ts` `modifyFeature`): shift the edited feature's scratch geometry into the world copy the user clicked (@daniel-freiermuth's `bestShift` search in #364 is the working pattern). There is **no `Select` / `Snap` / `Translate`** interaction in the app — `Modify` is the only interaction-internal hit-test.

**Standalone — WMS GetFeatureInfo.** `layer-wms-chart.component.ts` ~75 uses `getEventCoordinate` to build a server bbox request; a world offset there queries the wrong tile. Niche (WMS charts are rare) but the coordinate must be normalised before the request.

**One-off patch this fix replaces:**

- `fb-map.component.ts:533` — vessel-overlay `+360` shift that only fires while the dateline is in view (doesn't help the Keys case, or world −1 / +2). Remove it; the shared world offset supersedes it.

**Related longitude-wrapping code that is NOT in scope** (they handle *data* normalisation, not interaction placement — correct as-is; listed only so the shared concept is visible):

- `geoutils.ts:204` `normaliseCoords` — normalises clicked coordinates before storage (a data boundary; correct). Its in-place mutation / unbounded loops are separate quality nits, not this bug.
- `layer-currents-weather.component.ts:82` — weather-grid longitude handling, unrelated to interaction placement.

### Definition of done

- Panned one full world **east**, and again **west**, clicking every feature type opens its popup **on the icon**, not a world away.
- In Modify mode in a non-primary world copy, route vertices can be grabbed and dragged normally.
- None of the above is achieved by letting an out-of-range longitude reach a saved resource or the server: a route created or edited in a non-primary world saves with all coordinates in `[-180, 180]` (RFC 7946 §3.1.9).
- The three one-off patches above are removed, not left alongside the shared fix.
- Unit tests where the seam allows — the coordinate-emit helper and any Mercator offset math are testable; the OL interaction itself less so.
- `AGENTS.md` captures the render-space vs data-space concept and the forward rule (see *Documentation* below).

### Documentation (ship with the fix)

The whole point of fixing this at the source is that it stops recurring — but that only holds if the concept is written down where the next contributor (human or AI) will read it before touching map coordinates. **The fixing PR must update `AGENTS.md`** (in the same PR — it documents the very abstraction the fix introduces, so it is not a separate logical change). It should capture:

- **The two coordinate spaces.** *Render space* — unwrapped, world-copy-aware, OpenLayers-internal; used only for drawing, overlay placement, and hit-testing. *Data space* — canonical WGS84 in `[-180, 180]`; everything persisted, streamed, displayed, or read by another app. `toLonLat`/`fromLonLat` cross between them and silently normalise, which is right for data and wrong for placement.
- **The invariant.** Nothing written to the Signal K server may carry an out-of-range longitude (RFC 7946 §3.1.9). Normalise at every data boundary; keep the world offset in render space (Mercator) only.
- **The forward rule.** New code that converts a pointer event to a coordinate, places an overlay, or hit-tests must go through whatever shared mechanism this PR introduces (e.g. the coordinate-emit helper and the overlay-placement path) rather than calling `toLonLat`/`fromLonLat` ad hoc — so the pattern doesn't re-fragment into another round of one-off patches. Name the helper(s) and the single output choke so they're discoverable.

### Credit

Surfaced by @daniel-freiermuth in #364, which contains a working fix for much of this alongside its antimeridian rendering work. Related: #570.

joelkoz:
OpenLayers pans horizontally without limit and replicates features into every visible world copy, but Freeboard resolved every *interaction* to the primary world. `toLonLat()` normalises a click to `[-180,180]`, discarding which copy the user was looking at, so a popover opened a world away from the feature clicked, and route vertices could only be grabbed in the primary world. It had nothing to do with the antimeridian — a route in the Florida Keys reproduced it by panning one world over.

The fix introduces an explicit split between **render space** (EPSG:3857, world-copy-aware — placement and hit-testing) and **data space** (canonical WGS84 in `[-180,180]` — everything stored/streamed). A click carries a render-space **world offset** that is applied only in Mercator and never becomes a lon/lat, so nothing out-of-range can reach the server.

- `worldCopyOffset()` (util) derives the whole-world offset; map click events carry it as `worldOffset` while `lonlat` stays canonical.
- `ol-overlay` gains a `worldOffset` input and draws its (still canonical) `position` in that copy; a tracked vessel popover and the live measure label follow in the same copy. The dateline-only `+360` overlay hack this supersedes is removed.
- Route editing aligns the edit feature to the copy the user clicked **by its extent centre** — a route's own coordinates may be unwrapped past ±180, so the click's world alone is wrong — then shifts by a whole world so OL `Modify` (which hit-tests raw geometry in view space; its `wrapX` option only wraps the sketch overlay) matches where the user clicks. The shift is visually transparent under wrapX, adds no vertices, and normalises back to `[-180,180]` on save.

The render-space vs data-space rule and the shared helpers are documented in `AGENTS.md` so the pattern isn't re-fragmented.

Out of scope: WMS `GetFeatureInfo` (`layer-wms-chart`) has the same coordinate on a server bbox request; niche (WMS charts are rare) and a separate mechanism, noted as a follow-up.

### Test plan

- [x] `worldCopyOffset` unit tests (primary world, east/west copies, further copies, whole-world invariant, non-finite guards)
- [x] `npm run test:ci` — 281 tests pass; `npm run build:all` clean; prettier clean
- [x] Manual, against a live server with an antimeridian-crossing route: popups open on the icon one world east and west; route vertices editable in the primary world, one world east, one world west, and in the straddling (zoomed-out) case each half edits in its own copy; Florida Keys route edits normally
- [x] Invariant: a route created while panned **two worlds west** (and crossing the antimeridian) saved with every coordinate in `[-180,180]` — verified on disk

Closes #572

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map interactions near the international date line and across duplicated world views.
  * Kept popovers, measurement labels, and feature editing aligned with the world copy where the interaction occurred.
  * Preserved canonical geographic positions while displaying overlays in the correct rendered location.

* **Tests**
  * Added coverage for world-copy positioning, including adjacent and multiple wrapped worlds and invalid coordinate values.

* **Documentation**
  * Added guidance for separating geographic data coordinates from rendered map coordinates.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -98,6 +98,29 @@ live.)
   values omit it. Don't call `Convert` directly for display — `Convert` is the pure
   primitive for geometry and numeric math only. See
   [`docs/signalk/unit-preferences.md`](docs/signalk/unit-preferences.md).
+- **Map coordinates → keep render space and data space separate.** The map deals
+  with two coordinate spaces and must not confuse them:
+  - **Data space** — canonical WGS84 in `[-180, 180]`. Everything persisted to the
+    Signal K server, streamed, displayed, or read by another app. GeoJSON requires
+    it (RFC 7946 §3.1.9); a stored longitude of `197` is malformed and other
+    consumers may reject it. **Normalise at every data boundary** (`toLonLat`
+    already does this on the way in; the save path re-normalises on the way out).
+  - **Render space** — EPSG:3857 Mercator, world-copy-aware. OpenLayers pans
+    horizontally without limit and replicates geometry into every visible world
+    copy out to ±540° (see the *reading / exploring* lesson on the wrapping world),
+    so a click east/west of the primary world carries an x outside one world width.
+    This offset is what lets a popover or an edited vertex land in the copy the user
+    is actually looking at.
+  - **The rule.** When you convert a pointer event to a coordinate, place an
+    overlay, or hit-test/​edit a feature, carry the **world offset** in render space
+    and apply it in Mercator — never bake it into a lon/lat (that would leak an
+    out-of-range value toward storage). The shared machinery: `worldCopyOffset()`
+    (`ol/lib/util.ts`) computes the offset; map click events carry it as
+    `worldOffset`; `ol-overlay`'s `worldOffset` input places an overlay in that copy
+    while its `position` stays canonical; `Modify` edits shift the feature by a
+    whole-world offset (visually transparent under wrapX, normalised on save). Route
+    them through these rather than re-deriving with ad-hoc `toLonLat`/`fromLonLat`
+    or `±360` shifts — that fragmentation is exactly what #572 consolidated.
 - **Tests.** New behaviour needs tests where the test infrastructure supports it
   (`*.spec.ts`, run via `npm run test:ci`). Test behaviour, not implementation.
 
--- a/src/app/modules/map/fb-map.component.html
+++ b/src/app/modules/map/fb-map.component.html
@@ -511,7 +511,11 @@
 
   <!-- Distance & Bearing to -->
   @if (overlay().show && overlay().type === 'bearing_dist') {
-    <ol-overlay [position]="overlay().position" [positioning]="'top-center'">
+    <ol-overlay
+      [position]="overlay().position"
+      [worldOffset]="overlay().worldOffset ?? 0"
+      [positioning]="'top-center'"
+    >
       <ap-popover
         [title]="overlay().title"
         [canClose]="false"
@@ -539,7 +543,11 @@
 
   <!-- Popovers -->
   @if (overlay().show && !mapInteract.isModifying()) {
-    <ol-overlay [position]="overlay().position" [positioning]="'top-center'">
+    <ol-overlay
+      [position]="overlay().position"
+      [worldOffset]="overlay().worldOffset ?? 0"
+      [positioning]="'top-center'"
+    >
       @if (mapInteract.isMeasuring()) {
         <ap-popover
           [title]="overlay().title"
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -42,7 +42,7 @@ import { FreeboardOpenlayersModule } from 'src/app/modules/map/ol';
 import { CoordsPipe } from 'src/app/lib/pipes';
 
 import { computeDestinationPoint, getGreatCircleBearing } from 'geolib';
-import { toLonLat } from 'ol/proj';
+import { fromLonLat, toLonLat } from 'ol/proj';
 import { Style, Stroke, Fill } from 'ol/style';
 import { Collection, Feature } from 'ol';
 import { Feature as GeoJsonFeature } from 'geojson';
@@ -122,6 +122,7 @@ import { ScaleLine } from 'ol/control';
 import { Units } from 'ol/control/ScaleLine';
 import { DragBoxEvent } from 'ol/interaction/DragBox';
 import { MapService } from './ol/lib/map.service';
+import { worldCopyOffset } from './ol/lib/util';
 import { AppIconDef } from '../icons';
 import { LayerWindWeatherComponent } from './ol/lib/resources/layer-wind-weather.component';
 import { LayerCurrentsWeatherComponent } from './ol/lib/resources/layer-currents-weather.component';
@@ -285,6 +286,19 @@ export class FBMapComponent implements OnInit, OnDestroy {
 
   private saveTimer;
   private isDirty = false;
+  /**
+   * EPSG:3857-metre offset of the world copy the most recent click landed in.
+   * Passed to overlays so a popover is drawn in the copy the user clicked, not
+   * the primary world. Render-space only — never written to stored data (#572).
+   */
+  private clickWorldOffset = 0;
+  /**
+   * Raw EPSG:3857 x of the most recent click, world-copy-aware. Used to align an
+   * edited feature with the copy the user is looking at before Modify hit-tests
+   * it (a route's own coordinates may be unwrapped past ±180, so the click's
+   * world alone isn't enough — align by the feature's extent). See #572.
+   */
+  private clickMercX = 0;
 
   // Cursor position readout. A signal so the readout (and measure overlay)
   // refresh when pointer-move runs OUTSIDE the Angular zone (see MapComponent).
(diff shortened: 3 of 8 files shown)


joelkoz:
@coderabbitai pause

Issue #542, opened 2026-07-18
feat(routes): Add "Undo" functionality to route editing


joelkoz:
Support "Undo" operation for all route creation/editing.  Enable this via both a button on the popup route editing window(s), as well as common key presses like Ctrl-Z (windows)/Cmd-Z (macOS).

PR #582, opened 2026-07-25
feat(routes): unify the route draw and modify helper panels

### LINKED ISSUE #545: feat(routes): Unify route creation/editing windows
Currently FSK has the following route "helper" panels:

1. A popup that displays when drawing a new route for the first time (shows only distance)
2. A popup that displays when you click on an existing route (be it saved or unsaved) with "Modify", "Save", "Delete" as options
3. A popup that displays when you select "Modify" from 2

All three of those are different. `#1` and `#3` are SIMILAR, but differ in instructions and a "Finish" button.

When you click on an existing route and select "Modify", there is a Modify popup. That exact same window should be used when drawing (i.e. there should be a single Modify popup, and drawing is a form of modify).  If there are technical reasons these windows need to be different (e.g. they are different classes), they should be identical IN APPEARANCE.

Unify these popups look and feel so they are consistent.  Similar options, like "Save" should be in the same location. on all.  The "Undo" option (depends on #542 ) should swap places with "Modify" when you switch into editing mode.

The `#2` popup the user sees when you click on an unsaved route differs from when you click on a saved route. A saved route offers up "Points", which lets you re-order the points in the route in a popup waypoint list. There is no reason to not allow that on an unsaved route. Now it may be that the "Points" window only works on saved routes (and not the "route buffer" used for editing). If that is the case, we need to open a separate issue to refactor the "Points" dialog to work with the route buffer so it can be used by both saved and unsaved routes.

joelkoz:
Unifies the route creation/editing helpers into one consistent card (#545).

**Why.** Drawing a route, clicking a route, and modifying one each presented a
different helper. Draw and Modify in particular looked unrelated despite being
the same kind of operation.

**What.** Route Draw and Modify now share one docked popover card — the same
chrome as the feature popover — with distance at the top, the modify leg
read-out preserved, and `Finish` + `Cancel` actions in a fixed place. Drawing's
`Finish` completes the sketch via the OpenLayers draw interaction, so it matches
Modify's rather than relying on a double-click.

Before:
Screenshot 2026-07-24 at 10 06 58 PM

Screenshot 2026-07-25 at 12 57 54 PM

After:
Screenshot 2026-07-25 at 12 42 09 PM

Screenshot 2026-07-25 at 12 42 21 PM

**Behaviour change.** A route edit now commits directly from the card — `Finish`
saves, `Cancel` / the title-bar `✕` discards — replacing the old "Save route
changes?" confirm prompt. Non-route modifies (waypoint/note/region) keep their
confirm. Measure/box-select and other draws are unchanged.

Deferred to separate issues: the Undo↔Modify swap (depends on #542) and enabling
the Points reorder dialog on unsaved routes (needs the Points dialog to work
against the route buffer).

Closes #545

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1044,6 +1044,7 @@
           >
             <!-- MAP -->
             <fb-map
+              #fbMap
               [scaleUnits]="app.config.units.distance"
               [setFocus]="mapSetFocus()"
               [mapCenter]="mapCenter()"
@@ -1124,7 +1125,11 @@
 
           <!-- Interaction Help Container Panel -->
           @if (isInteracting()) {
-            <fb-interact-help (cancel)="closeInteraction()"> </fb-interact-help>
+            <fb-interact-help
+              (finish)="handleInteractionFinish()"
+              (cancel)="closeInteraction()"
+            >
+            </fb-interact-help>
           }
 
           <!-- *** Nav data panel ***-->
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -181,6 +181,7 @@ interface DrawEndEvent {
 })
 export class AppComponent {
   @ViewChild('sideright', { static: false }) sideright;
+  @ViewChild('fbMap', { static: false }) fbMap: FBMapComponent;
 
   protected navDataPanel = signal<{
     show: boolean;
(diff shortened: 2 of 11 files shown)


joelkoz:
@coderabbitai pause

PR #595, opened 2026-07-26
feat(charts): order chart list to match the chosen layer order

### LINKED ISSUE #550: feat(charts): order the chart list to match the user-chosen layer order
## What
The Charts panel has two screens: the main list (toggle visibility, opacity, image
adjustment, properties) and a separate "Re-order" screen ("Chart Order") where you
drag charts into your preferred layer order. The chosen order is applied to the map,
but the main list ignores it and stays in fetch order. The main list should present
charts in the user-chosen order, matching the Re-order screen.

## Current behaviour
- The main list renders `filteredList()`, sourced from `listFromServer` +
  `appendOSM` (server/fetch order). It never consults `chartOrder`
  (`src/app/modules/skresources/components/charts/chartlist.ts:122-127`; template
  `chartlist.html:133`).
- The Re-order screen shows charts arranged by `chartOrder` (top layer first) and,
  on drop, writes `selections.chartOrder` and calls `chartReorder()`, which only
  re-stacks the map layers
  (`src/app/modules/skresources/components/charts/chart-layers.component.ts:120-136`;
  `arrangeChartLayers` / `chartReorder` in
  `src/app/modules/skresources/resources.service.ts:720-757`).
- Result: reordering charts changes the map but not the list, so the list and the
  Re-order screen disagree.

## Expected
The main chart list is presented in the same order as the Re-order (Chart Order)
screen — top layer first — so what you see in the list matches the layer order you
chose. New charts not yet in `chartOrder` fall in at a stable position (they're
already appended to `chartOrder` by `arrangeChartLayers`).

## Suggested direction
Order the list off `selections.chartOrder` (the same ordering the Re-order screen
uses) rather than fetch order, and keep the presentation direction consistent
between the two screens.

## Done when
After re-ordering charts, the main chart list shows them in that order, matching the
Re-order screen; toggling visibility / opacity / properties is unaffected.

---
🤖 Created with [Claude Code](https://claude.com/claude-code)

joelkoz:
## What
The main Charts list is now presented in the user-chosen layer order — top
layer first — matching the Re-order (Chart Order) screen. Previously the list
sorted alphabetically by name and ignored `chartOrder`, so the list and the
Re-order screen disagreed after a re-order.

## How
- `doFilter()` orders the list through the same
  `skres.arrangeChartLayers(fullList).reverse()` transform the Re-order screen
  uses, then applies the text / in-view filters preserving that order (removing
  the `localeCompare` name sort). Reusing `arrangeChartLayers` guarantees the two
  screens stay in agreement.
- Closing the Re-order screen re-runs `doFilter()`, so a re-order is reflected in
  the list immediately on return.

Toggling visibility / opacity / properties is unaffected.

## Tests
`chartlist.spec.ts` covers top-first ordering, that it is no longer alphabetical,
that filtering preserves layer order, and that closing the Re-order screen
re-applies a changed order.

Closes #550

## Summary by CodeRabbit

* **Bug Fixes**
  * Chart lists now follow the configured layer order instead of alphabetical ordering.
  * Filtering by chart name or map extent preserves the selected layer order.
  * Updated layer ordering is reflected immediately after leaving the chart reorder screen.

* **Tests**
  * Added coverage for ordering, text filtering, extent filtering, and reordered chart results.

### DIFF
--- a/src/app/modules/skresources/components/charts/chartlist.spec.ts
+++ b/src/app/modules/skresources/components/charts/chartlist.spec.ts
@@ -12,13 +12,11 @@ import { FBMapInteractService } from 'src/app/modules/map/fbmap-interact.service
 import type { FBChart, FBCharts } from 'src/app/types';
 
 /**
- * The main chart list must present charts in the user-chosen layer order — top
- * layer first — the same ordering the Re-order (Chart Order) screen uses, rather
- * than the previous alphabetical-by-name sort (#550). The re-order screen derives
- * its order from `skres.arrangeChartLayers(list).reverse()`; `doFilter()` now
- * applies the identical transform, so the two screens agree. These tests drive
- * `doFilter()` with a stubbed `arrangeChartLayers` and assert the observable
- * order of the `filteredList` signal.
+ * The main chart list presents charts in the user-chosen layer order — top layer
+ * first — matching the Re-order (Chart Order) screen. Both derive that order from
+ * `skres.arrangeChartLayers(list).reverse()`, so the two screens agree. These
+ * tests drive `doFilter()` with a stubbed `arrangeChartLayers` and assert the
+ * observable order of the `filteredList` signal.
  *
  * `TestBed.inject` gives a real injection context for the constructor effects
  * without rendering the template.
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai pause

PR #601, opened 2026-07-26
feat(map): combine center and follow vessel into one button

### LINKED ISSUE #541: Streamline center & follow vessel buttons
Now there are two buttons, one for Follow Vessel and another for Center Vessel:

Image

I think we could save some screen real estate, especially on the mobile, with just a single button

Image

The button could alternate between
- off
- keep vessel centered
- lookahead / offset vessel center

Any manual movement of the map switches to off.

The button should give a visual indication of the mode, to give immediate feedback when you press it.

A further optimisation would be that when off pressing the button switches to last chosen follow vessel mode, either centered or lookahead. This way the offset vessel center setting would be just a setting for the offset amount, not a global switch to always use it like it is now. A few taps of the button would allow you to switch between the modes. I imagine in practise you'd mostly alternate between manually panning the chart (switches to off) and a single button press (switches to your favorite follow vessel mode).

joelkoz:
Follow Vessel and Center Vessel occupied two toolbar slots for what is really one
intent, which costs real estate on a phone. They are now one button.

**Button** — a tap centres the vessel and turns follow mode on (highlighted); a
tap while on turns it off and leaves the chart where it is. The configured centre
offset is held off for two seconds after follow mode starts, so a second tap
straight away leaves the vessel dead centre — that is how "just centre it"
survives the merge into one button.

**Panning while following** — the old *Lock Follow Vessel* checkbox becomes a
*Map Pan When Following* choice:

- **Set Follow Offset** — follow stays on and the pan sets the offset by eye
- **Exit Follow Mode** — panning releases follow (what unlocked used to do)
- **Do nothing** — vessel returns to position on the next fix (what locked used
  to do); keeps an accidental screen tap from disturbing anything

Existing configs migrate on load: locked → *Set Follow Offset* (it meant "panning
must not drop me out of follow mode"), unlocked or absent → *Exit Follow Mode*.

**Offset Vessel Center** is now entered directly as a signed whole percentage
(±90) instead of chosen from a preset list, so panning has somewhere to write to
and a look-behind is expressible. Stored fractions migrate ×100; a non-integer is
the reliable tell for a legacy value.

The offset is a single along-course value, so a pan off the course line
contributes only its along-course component and the vessel springs back onto the
course line.

Releasing follow mode no longer re-centres the vessel — under one button that
would undo the pan that just released it.

Closes #541

## Summary by CodeRabbit

* **New Features**
  * Added configurable map panning behavior while following a vessel: set a follow offset, exit follow mode, or do nothing.
  * Replaced the offset preset selector with a percentage input, including validation and limits.
  * Panning while following can now automatically calculate and save the vessel’s follow offset.
  * Added smoother follow-mode centering to prevent abrupt snap-backs after map interactions.

* **Documentation**
  * Updated tooltips and help content to explain vessel centering, follow offsets, and map-panning options.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -543,27 +543,13 @@
                     [matTooltip]="
                       app.uiConfig().mapMove
                         ? 'Turn off Follow Vessel'
-                        : 'Follow Vessel'
+                        : 'Center & Follow Vessel'
                     "
                     matTooltipPosition="right"
                   >
                     <mat-icon class="ob" svgIcon="cent-iec"></mat-icon>
                   </button>
                 </div>
-                @if (!app.uiConfig().mapMove) {
-                  <div class="buttonPanelItem">
-                    <button
-                      class="button-toolbar"
-                      mat-mini-fab
-                      (click)="centerVessel()"
-                      [disabled]="!app.data.vessels.showSelf"
-                      matTooltip="Center Vessel"
-                      matTooltipPosition="right"
-                    >
-                      <mat-icon> center_focus_strong </mat-icon>
-                    </button>
-                  </div>
-                }
                 <div class="buttonPanelItem">
                   <button
                     class="button-toolbar"
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1481,14 +1481,6 @@ export class AppComponent {
     }
   }
 
-  /**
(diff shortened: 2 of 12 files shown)


joelkoz:
@coderabbitai pause

PR #604, opened 2026-07-26
feat(features): show a feature's newest changes above its instructions

### LINKED ISSUE #603: Feature Browser: surface a feature's newest changes above the body, not only in the History table
## Problem

In the Feature Browser's details pane, a feature's changes are only visible in the
**History** table at the very bottom, below the whole instruction body. A user skimming
"what's new" has to read past (or scroll through) the instructions to reach it, and many
won't scroll at all.

The instructions and the change list are competing for the top of the pane. Moving the
History table up would just invert the problem — then the instructions need scrolling.

## Proposed solution

Render a short, generated **"New in x.y"** strip between the meta chips and the body. The
full History table stays where it is, unchanged.

```
┌─ Route Planning ─────────────────────────┐
│  [ lead screenshot ]                     │
│  Route Planning                          │
│  ‹Navigation›  ‹Updated in 3.1›          │
│                                          │
│  ✨ New in 3.1                           │  ← generated from the ledger
│   · Insert a waypoint into a leg         │     titles only; no dates, no PR links
│   · Reverse a route                      │
│   · Show leg bearing while drawing       │
│   +2 more ↓                              │  ← anchors to the History table
│                                          │
│  To create a route, tap …                │  ← instructions stay near the top
│  …                                       │
│                                          │
│  History                                 │  ← unchanged, still at the bottom
└──────────────────────────────────────────┘
```

Nothing is written into any `features/*.md`. The strip is computed from the ledger rows
already in `features/changelog.json`, so it cannot drift from the History table, needs no
authoring step, and **ages out by itself** — when 3.2 ships, the 3.1 strip stops rendering
with no action taken.

## Proposed behaviour

The strip covers **the current release series** — everything since the last official
release. On a 3.1 build it shows what landed after 3.0, and keeps showing it until 3.2
ships. It is the same for every user; nothing is tracked per-user or per-page.

- `currentSeries` = the **major.minor** of the newest `since` in the ledger
  (`compareSince` to find it, then take the first two components).
- A feature renders the strip when it has any row in `currentSeries` (i.e. `since`
  matching `3.1.*`), **or** any unreleased row (so dev builds and betas aren't blank).
- Rows included: every row of the feature in `currentSeries`, plus any with no `since`.
- Heading: **"New in 3.1"** when the feature's row in that group is its `new` row,
  **"Updated in 3.1"** otherwise — matching the badge already shown in the list. With only
  unreleased rows: **"Unreleased"**. The heading names the series (`3.1`), not the exact
  patch version.
- Titles rendered via the existing `stripTypePrefix`; capped at 3, with `+N more` linking
  down to the History table.

**Why the series and not the exact version.** Matching `since` exactly against the newest
release breaks on patch releases: if 3.1.1 ships carrying a single enhancement to one
feature, that row becomes the newest `since` and every 3.1.0 strip disappears mid-series.
Grouping by `3.1.x` keeps the strip stable for the whole release series, which is the
intended behaviour.

## Out of scope

- **Gating on unseen events instead of the release.** `whats-new.ts` already keys
  seen-state on `<feature-id>:<pr>`, so the strip could read *"New since you last looked"*.
  **Rejected**, not merely deferred: the strip must not disappear once someone has looked
  at it. A change is "new in 3.1" for the whole life of 3.1, however many times the browser
  is opened — and that also keeps it identical for every user on that version, matching the
  release notes. (The existing lightbulb indicator is unaffected; it stays per-user.)
- Any change to the History table itself.

## Implementation notes

| File | Change |
|---|---|
| `src/app/modules/features/feature-corpus.ts` | add `recentChanges` to `CompiledFeature`; derive `currentSeries` and compute the group + label in `compileCorpus` |
| `src/app/modules/features/feature-corpus.spec.ts` | cover the gate, the unreleased case, the cap, a feature *not* in the current series, and **a patch release within the series not hiding earlier rows** |
| `src/app/modules/features/feature-browser-dialog.html` / `.css` | render the strip between `.meta` and `.body` |
| `features/route-planning.md` | delete the hand-written `## New in 3.1` section this replaces |

joelkoz:
A feature's changes were only visible in the History table at the very bottom of
the details pane, below the whole instruction body — so anyone skimming "what's
new" had to read past the instructions to reach it, and most never scrolled that
far. Moving the History table up would only invert the problem.

Instead the details pane now carries a short generated strip between the meta
chips and the body: the feature's changes in the **current release series**,
titles only, capped at three with a `+N more` affordance that scrolls to the full
history. The History table is untouched.

The strip is computed from the ledger rows already in `features/changelog.json`,
so nothing is written into any `features/*.md`, it cannot drift from the history
table, and it ages out on its own — once 3.2 ships, the 3.1 strip stops rendering
with no action taken. This replaces the hand-written `## New in 3.1` section in
`route-planning.md`, which would have rotted for exactly that reason.

**Scoped by `major.minor`, not the exact `since`.** "New in 3.1" has to mean
everything after 3.0 for the whole life of 3.1. Matching the newest release
exactly would break on a patch: one enhancement stamped `v3.1.1` becomes the
newest release and every 3.1.0 strip disappears mid-series. There's a regression
test for that case.

**Unreleased rows form their own group**, labelled `Unreleased`, rather than
joining the current series. Folding them in would announce not-yet-shipped work
under the shipped release's name — against the real ledger today that would print
"New in 3.0" over five changes headed for 3.1. Released builds never carry
unstamped rows, so this branch only affects dev and beta builds, where it keeps
the strip from being blank.

Verified in the running app (light and dark) against the real corpus.

Closes #603

## Summary by CodeRabbit

- **New Features**
  - Added a “What’s New” strip in the feature details pane, shown when recent updates exist.
  - Displays a capped list (up to three) of recent change titles, labeled by release context (including “Unreleased” when applicable).
  - Includes a “+N more” control that smoothly jumps to the full PR history.

- **Documentation**
  - Removed the outdated “New in 3.1” route-planning bullet list (route-instruction narrative remains).

### DIFF
--- a/features/route-planning.md
+++ b/features/route-planning.md
@@ -9,15 +9,6 @@ category: Routes
 Draw a route on the chart, review and adjust it, then save it for later —
 or skip saving and just use it right away.
 
-## New in 3.1
-
-- **One editing card** for both drawing and modifying, with the commit and
-  **Cancel** actions always in the same place.
-- **Undo** while drawing or editing — an Undo button, or Ctrl-Z / ⌘-Z.
-- **Extend a saved route** by clicking open water, without leaving Modify.
-- **Reorder points on an unsaved draft**, not just on saved routes.
-- **Hide** a route from the chart directly from its popover.
-
 **To draw a route**, open the edit menu (pencil icon) and choose **Draw
 Route**. Tap each point along your intended route. A small editing card
 shows the running distance and a **Finish** button — tap **Finish** (or
--- a/src/app/modules/features/feature-browser-dialog.css
+++ b/src/app/modules/features/feature-browser-dialog.css
@@ -118,6 +118,51 @@ figure.shot figcaption {
   opacity: 0.7;
 }
 
+/* "New in x.y" strip: the current release series' changes, above the body so a
+   reader skimming what's new never has to reach the history table below it.
+   Colours come from the primary-container tokens, which the dark-theme block at
+   the foot of this file redefines — so this follows the theme for free. */
+.whats-new {
+  display: flow-root; /* narrow beside the lead figure, not forced below it */
+  margin: 0 0 14px;
+  padding: 8px 12px;
+  border-radius: 8px;
+  border-left: 3px solid var(--mat-sys-primary, #415f91);
+  background: var(--mat-sys-primary-container, #d7e2ff);
+  color: var(--mat-sys-on-primary-container, #274777);
+}
+.whats-new-title {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  font-size: 0.72rem;
+  font-weight: 600;
+  letter-spacing: 0.02em;
+  text-transform: uppercase;
+}
+.whats-new-title mat-icon {
+  font-size: 16px;
+  width: 16px;
+  height: 16px;
+}
+.whats-new ul {
+  margin: 5px 0 0;
+  padding-left: 20px;
+  font-size: 0.85rem;
+  line-height: 1.5;
+}
+.whats-new-more {
+  margin-top: 5px;
+  padding: 0;
+  border: 0;
+  background: none;
+  font: inherit;
+  font-size: 0.8rem;
+  color: inherit;
+  text-decoration: underline;
+  cursor: pointer;
+}
+
 /* PR history table in the details pane (most recent first). */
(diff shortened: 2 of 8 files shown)


joelkoz:
@coderabbitai review

PR #607, opened 2026-07-27
feat(plotterext): emit map.view when the chart viewport changes

### LINKED ISSUE #565: feat(plotterext): Add events to notify listeners when the chart viewport has changed (panned and/or zoomed)
Some extensions want to know what the current chart viewport is (or if it has changed).  Adding events that announce a change prevents the need for extensions to poll the viewport.

joelkoz:
Implements the `map.view` host event in the reference host, per the contract
merged in #606.

An extension whose work depends on what the user is looking at — search-in-view,
area downloads, an overlay that refetches for the visible box — previously had no
way to learn the viewport moved and had to poll `map.getView`.

`onMapMoveEnd` (the OpenLayers `moveend` handler) already writes all three view
values, so a reactive `effect` on `mapExtent` gives one emission per *settled*
pan/zoom rather than per frame, and it is origin-transparent for free: OL fires
`moveend` for the user dragging, for Freeboard recentring on the vessel, and for
an extension's own `map.center` / `map.fitBounds`.

Two details worth a reviewer's eye:

- `map.getView` and the event now build their payload from one `mapView()`
  helper, so the method and the event cannot drift apart — the spec requires them
  to carry the same shape.
- `emitMapViewChange` compares by **value**. OL hands over a fresh extent array
  on every `moveend`, so the signal is re-assigned even when the view came back to
  where it started; a reference check would emit spurious events.

`plotterext.embedding-host.spec.ts` needed a deeper `AppFacade` stub: its async
tests flush the constructor effects, and the new one reads `config.map`. That is
the trap already documented in `DEV-LESSONS-LEARNED.md` under *When testing*.

Requires `signalk-plotterext-bus` ^0.11.0 (published), which adds the `MapView`
type this imports.

Closes #565

## Summary by CodeRabbit

* **New Features**
  * Plotter Extensions can now access the chart viewport and request map movements.
  * Added `map.view` events containing the current center, zoom, and bounds after the view settles.
  * Repeated events are suppressed when the viewport has not changed.

* **Documentation**
  * Documented viewport access, movement requests, and map view event behavior for extensions.

* **Tests**
  * Added coverage for viewport events, unchanged views, and zoom-only changes.

### DIFF
--- a/docs/freeboard/freeboard-plotter-ext-support.md
+++ b/docs/freeboard/freeboard-plotter-ext-support.md
@@ -245,6 +245,37 @@ neither field present), `nightMode.notSupported`.
 | `src/app/modules/plotterext/plotterext.service.ts` | binds the handlers (`readNightMode`/`applyNightMode`) and emits `nightMode.changed` (`emitNightModeChange`) |
 | `src/app/modules/skstream/skstream.facade.ts` | `selfNightMode` signal + `refreshSelfNightMode()` |
 
+## The `map` capability
+
+`map` exposes the chart viewport: `map.getView` reads it, `map.center` /
+`map.fitBounds` drive it, and `map.view` follows it.
+
+Moves are routed through Freeboard's own centering path
+(`AppFacade.mapMoveRequest` → `AppComponent` effect → `centerAndZoom`), never by
+reaching into the OpenLayers view directly — driving the OL view bypasses the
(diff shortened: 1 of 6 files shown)


joelkoz:
@coderabbitai pause

joelkoz:
@coderabbitai resume

PR #609, opened 2026-07-27
fix(routes): report the leg ending at the dragged vertex when modifying a route

### LINKED ISSUE #581: bug(routes): Route modify "Leg" read-out shows the last segment instead of the point being dragged
When modifying a route, the editing helper shows a **Leg** read-out (distance +
bearing of one segment of the route). It currently always shows the **last**
segment: the segment index defaults to `-1` and `measurements.component.ts`
`calc()` uses `coords.slice(-2)`. It also only recomputes when a drag is released
(`onModifyEnd`), not live during the drag.

**Result:** dragging a middle vertex leaves the Leg unchanged (the last segment
didn't move); only dragging the end point (or the second-to-last) updates it.

**Expected:** the Leg should describe the point you just moved — the segment from
the dragged vertex to the previous point — and, for the first point, the initial
segment.

**Sketch:** on modify-end, diff the pre/post coordinates to find the moved vertex
index, then set `measurement().index` to it (clamped so the last point maps to the
final segment). `fb-map.component.ts` `applyModifyEnd()` already sets
`measurementCoords`; it just needs to set the index too.

Surfaced by #545 — the unified route helper makes this read-out prominent.

joelkoz:
The **Leg** read-out shown while modifying a route always described the *last* segment, whatever the user was actually dragging. `measurement().index` was set to `-1` when the interaction started and never updated, and `-1` means "last leg" — so moving a middle vertex left the read-out unchanged and it appeared to work only when the end point moved.

A Modify operation already gives both sides of the geometry, so the touched vertex is just the first index at which they differ — for a move the vertex itself, for an insert the new vertex, and for a delete the vertex that closed the gap. The read-out reports the leg *ending* at that vertex, i.e. from the previous point to the one dragged; the first point has no preceding leg, so dragging it reports the leg leaving it.

That decision is a pure function (`route-modify-leg.ts`) called from the three paths that update measurement coords during a route modify — a vertex move/insert/delete, an undo, and a route extend — so undo and extend keep the read-out in step rather than leaving a stale index behind.

Surfaced by #545, which makes the read-out prominent.

Closes #581

### Test plan

- [x] `npm run test:ci` — 404 passing, incl. 12 new cases covering move / insert / delete / append / first-point / end-point / no-op
- [x] Red→green verified: with the helper stubbed to the old `-1` behaviour, 10 of the 12 new cases fail
- [x] `npm run build:all` clean
- [x] Manually exercised against a local Signal K server: drag middle / first / end vertex, insert, delete, undo, extend

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved measurement leg selection when route vertices are moved, undone, inserted, deleted, or extended.
  * Corrected the displayed leg and vertex readout for edits at the start, middle, and end of a route.
  * Added safeguards to keep selected leg indexes within valid route bounds.

* **Tests**
  * Added coverage for route editing scenarios, including vertex movement, insertion, deletion, extension, and unchanged routes.

### DIFF
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -131,6 +131,7 @@ import {
   RoutePointMeta,
   WORLD_WIDTH_3857
 } from './route-extend';
+import { modifiedLegIndex } from './route-modify-leg';
 import { AppIconDef } from '../icons';
 import { LayerWindWeatherComponent } from './ol/lib/resources/layer-wind-weather.component';
 import { LayerCurrentsWeatherComponent } from './ol/lib/resources/layer-currents-weather.component';
@@ -1065,6 +1066,10 @@ export class FBMapComponent implements OnInit, OnDestroy {
     if (!f) {
       return;
     }
+    // Capture the geometry before it is replaced: diffed against the restored
+    // coordinates it tells the Leg read-out which vertex the undo puts back
+    // (#581).
+    const undone = f.getGeometry().getCoordinates() as Position[];
     // setCoordinates fires the feature's 'change' event; the active OL Modify
     // interaction listens for it (handleFeatureChange_) and rebuilds its
     // segment/vertex index whenever the geometry changes outside its own drag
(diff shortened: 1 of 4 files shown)


joelkoz:
@coderabbitai pause

PR #612, opened 2026-07-27
fix(routes): don't append an end point on the click that deletes a vertex

### LINKED ISSUE #608: fix(routes): click-and-hold vertex delete also appends an end point
Deleting a route vertex with click-and-hold also appends a new end point to the route: the same release that completes the delete is taken as an "extend from the end" click, so the end of the route jumps to where the deleted point was.

Introduced by the route-extend behaviour added in #598.

### Steps to reproduce

1. Modify a route of 4+ points.
2. Click and hold (1500 ms) on an interior vertex to delete it.
3. Release.

**Expected:** the vertex is removed and nothing else changes.

**Actual:** the vertex is removed *and* a new end point is appended at the position just deleted, so the route's end jumps there.

Affects mouse and touch alike — `pointerDownHandler` in `ol/lib/map.component.ts` arms the long-press timer for any pointer type, so click-and-hold sets `vertexDeleteOnRelease` the same way tap-hold does.

### Cause

The delete and the extend are driven by the *same* pointer release:

- `vertexDeleteCondition` (`ol/lib/interactions/interaction-modify.component.ts`) consumes `vertexDeleteOnRelease` on `pointerup` / `singleclick` / `click` and deletes the grabbed vertex.
- `extendRouteInModify` (`map/fb-map.component.ts`) runs from the map's single-click handler and appends an end point when the click *misses* the route.

Its only guard is a `getFeaturesAtPixel` hit-test against the route's **current** geometry. By the time it runs the vertex is already gone, so the line has snapped away from the pixel that was clicked — the hit-test misses and the click is classified as open water.

### Suggested fix

Make the extend handler ignore a release that has just performed a delete, rather than relying on the hit-test alone — e.g. have `vertexDeleteCondition` record that it consumed a delete for this gesture (a flag on the map, cleared on the next `pointerdown` alongside `vertexDeleteOnRelease`) and have `extendRouteInModify` bail when it is set.

A regression test belongs with the fix: the extend decision is worth isolating into a pure predicate the way #598's geometry work went into `route-extend.ts`.

joelkoz:
Deleting a route vertex also appended a new end point to the route, so the end
of the route jumped to where the deleted point was (#608, a regression from the
route-extend behaviour in #598).

The delete and the extend ride the same pointer release. `extendRouteInModify`
guarded only on a `getFeaturesAtPixel` hit-test against the route's *current*
geometry — but by the time the resulting `singleclick` is handled the vertex is
already gone and the line has snapped away from the clicked pixel, so the
hit-test misses and the click is classified as open water.

Both delete paths are affected: the mid-hold `removePoint()` in
`tryDeleteHeldVertexOnHold` (which clears `vertexDeleteOnRelease` on success, so
nothing marked the release at all) and the delete-on-release consumed by
`vertexDeleteCondition` — the latter covering Ctrl-Click as well as tap-hold.
Both now mark the delete on the map, and the extend decision moves into a pure
`shouldExtendRoute()` predicate in `route-extend.ts` so it can be tested directly.

The marker deliberately outlives the gesture that set it. OpenLayers delays
`singleclick` by 250 ms and nothing cancels it at `pointerdown`, so clearing the
marker there loses a race: a fast follow-up press clears it before the pending
click is handled and the route is extended anyway. It is retired where a gesture
resolves instead — `singleclick`, `dblclick` (which cancels the pending
`singleclick`), and a drag (`dragging_` is sticky per gesture and gates click
emulation, so a gesture that has dragged emits no click at all).

Tested on a local dev server: long-press delete removes the vertex and nothing
else changes; a click in open water still extends the route.

Closes #608

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented accidental route extension after vertex deletion gestures.
  * Improved tap-hold, Ctrl-click, and long-press deletion handling so follow-up clicks/undo behavior stay in sync.
  * Ensured vertex-deletion suppression markers are properly retired or cleared across single clicks, double-clicks, and drag gestures.
* **Tests**
  * Added/expanded unit coverage for route-extension decision logic and the vertex-deleted gesture-marker behavior.

### DIFF
--- a/src/app/modules/map/fb-map.component.ts
+++ b/src/app/modules/map/fb-map.component.ts
@@ -129,8 +129,10 @@ import { LineString as OLLineString } from 'ol/geom';
 import {
   extendRouteAtClick,
   RoutePointMeta,
+  shouldExtendRoute,
   WORLD_WIDTH_3857
 } from './route-extend';
+import { vertexDeleted } from './ol/lib/vertex-delete';
 import { modifiedLegIndex } from './route-modify-leg';
 import { AppIconDef } from '../icons';
 import { LayerWindWeatherComponent } from './ol/lib/resources/layer-wind-weather.component';
@@ -1321,8 +1323,9 @@ export class FBMapComponent implements OnInit, OnDestroy {
    * the new end point (issue #549) — so a route can be lengthened the same way
    * it is drawn, without leaving modify mode. A click *on* the route belongs to
    * OL Modify (move / insert / delete a vertex), so only clicks that miss it
-   * extend the route; the state kept in sync here mirrors applyModifyEnd so
-   * Undo, Finish and Cancel all see the new point.
+   * extend the route — and never the click that completes a vertex delete
+   * (#608); the state kept in sync here mirrors applyModifyEnd so Undo, Finish
+   * and Cancel all see the new point.
    */
   private extendRouteInModify(e: FBClickEvent) {
     const f = this.mapInteract.draw.features?.getArray()[0] as Feature;
@@ -1341,7 +1344,14 @@ export class FBMapComponent implements OnInit, OnDestroy {
       .getMap()
       .getFeaturesAtPixel(e.pixel, { hitTolerance: tol })
       .some((ft) => ft.getId() === f.getId());
-    if (onRoute) {
+    // A delete has already moved the line away from the clicked pixel, so the
+    // hit-test above cannot recognise the release that completed it (#608).
+    if (
+      !shouldExtendRoute({
+        onRoute,
+        vertexDeleted: vertexDeleted(this.olMap.getMap())
+      })
+    ) {
       return;
     }
     // Metadata may not be seeded yet if this is the first op of the session (no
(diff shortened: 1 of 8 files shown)


joelkoz:
Good catch — confirmed and fixed in 5a11eaf9.

You are right that the `pointerdrag` clear was not scoped to the gesture that set the marker: delete a vertex and then immediately start panning the chart, and that unrelated drag retires the marker while the delete's `singleclick` is still pending — so the pending click appends an end point. Same class of bug as the `pointerdown` clear, one level down.

Per-gesture identity via OL is not available at the point the marker is set: `click`/`singleclick`/`dblclick` all carry `MapBrowserEventHandler`'s `down_` clone as `originalEvent`, but the tap-hold delete happens mid-gesture (before any release) and the delete-on-release path sees the real `pointerup`, not that clone. `down_` is also a fresh `new PointerEvent(...)`, so a tag put on the DOM `pointerdown` does not survive into it.

So the gesture id is stamped on our side instead: `startPointerGesture()` increments a monotonic counter at `pointerdown`, `markVertexDeleted()` records the current value, and `clearVertexDeletedOnDrag()` clears only when the dragging gesture is the one that set the marker. `singleclick`/`dblclick` still clear unconditionally — those *are* the marker's own click resolving.

Covered by two specs in `vertex-delete.spec.ts`: the deleting gesture's own drag retires the marker, and a later gesture's drag leaves it intact.

joelkoz:
@coderabbitai review

joelkoz:
@coderabbitai review

joelkoz:
@coderabbitai pause

PR #616, opened 2026-07-27
feat(map): offset the vessel abeam as well as ahead when following

### LINKED ISSUE #615: Follow offset should be settable in both axes, not just along course
Following up on feedback from #541.

#541 shipped a single **Offset Vessel Center** value, measured along the
vessel's course, and panning the chart under *Set Follow Offset* writes it. That
means a pan is projected onto the course line and its sideways component is
discarded — drag the vessel down and to one side, and it springs back onto the
course line.

The other chart plotters that inspired the drag-to-set gesture keep the offset
in **both** axes, so the vessel stays where you put it. That is what people
expect when they drag: the boat should stay under their finger, not slide back.

### Proposed change

Keep one offset concept, expressed as two values instead of one:

- **ahead / astern** — the existing value, along the vessel's course
- **abeam** — new, perpendicular to the course, so the vessel can sit off to one
  side

Both remain a **percentage of the distance from the centre of the screen to its
edge** in their own direction, so neither can push the vessel off screen.
Keeping the offset relative to the course (rather than fixed to the screen)
matters for north-up: the look-ahead has to follow the boat around as it turns,
otherwise heading south would show you the water behind you.

Panning under *Set Follow Offset* sets both values at once.

### Settings

**Offset Vessel Center** becomes a group label over two inputs in the space the
single percentage occupies today, rather than a single field.

### Compatibility

The existing stored value carries over unchanged as the ahead/astern offset; the
new abeam offset defaults to zero, so nobody's current view moves until they
choose to change it.

joelkoz:
Panning under *Set Follow Offset* projected the drag onto the course line and
threw away the sideways part, so dragging the vessel off to one side sprang it
back. The plotters that inspired the gesture keep the offset in both axes, and
that is what people expect when they drag — the boat should stay where they put
it.

**Offset Vessel Center** is now a group over two inputs, **Ahead** and
**Abeam**, and a pan sets both. Each stays a percentage of the distance from the
centre of the screen to its edge.

Both axes are measured in the vessel's frame rather than the screen's, so the
offset rotates with the boat. That matters in north-up: a screen-fixed
look-ahead would show a vessel heading south the water it had already passed.

Per-axis clamps alone do not keep the vessel visible — 90% ahead combined with
90% abeam lands diagonally outside the viewport corner — so the resolved offset
is scaled back uniformly when it would push the vessel off screen. Offsets on a
single axis never reach that point and are unchanged.

The stored value carries over as the ahead offset and the new abeam offset
defaults to zero, so no existing view moves.

Closes #615

## Summary by CodeRabbit

- **New Features**
  - Map centering offsets now support independent **Ahead** and **Abeam** positioning.
  - Follow mode now calculates offsets using the viewport size and map rotation, supporting diagonal and rotated views.
  - Settings added an **Abeam** offset input with validation, and it defaults to **0**.

- **Bug Fixes**
  - Improved offset math keeps the vessel correctly positioned on-screen, including corner and turn scenarios.
  - Panning now preserves both Ahead and Abeam components and returns consistent round-trip results.

- **Documentation**
  - Updated Map settings help to explain separate Ahead/Abeam behavior and follow-mode impact.

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -160,6 +160,9 @@ export function cleanConfig(
   } else {
     settings.map.centerOffset = clampCenterOffset(settings.map.centerOffset);
   }
+  settings.map.centerOffsetAbeam = clampCenterOffset(
+    settings.map.centerOffsetAbeam ?? 0
+  );
   if (typeof settings.map.doubleClickZoom === 'undefined') {
     settings.map.doubleClickZoom = false;
   }
@@ -530,6 +533,7 @@ export function defaultConfig(): IAppConfig {
       doubleClickZoom: false, // true=zoom
       overZoomTiles: true, // keep tiles visible beyond chart max zoom
       centerOffset: 0,
+      centerOffsetAbeam: 0,
       s57Options: {
         graphicsStyle: 'Paper',
         boundaries: 'Plain',
--- a/src/app/app.facade.ts
+++ b/src/app/app.facade.ts
@@ -11,8 +11,9 @@ import { Subject } from 'rxjs';
 import { InfoService, IndexedDB, AppInfoDef } from './lib/services';
 import { isTrackShown, toggleTrackSelection } from './lib/vessel-track';
 import {
+  CenterOffset,
   centerOffsetFromPan,
-  edgeDistanceInDirection,
+  MapViewport,
   mapCenterForOffset
 } from './lib/follow-offset';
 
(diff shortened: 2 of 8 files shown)


joelkoz:
@coderabbitai review

joelkoz:
Addressed the actionable finding: `centerOffsetFromPan` now runs its result through the same corner constraint the renderer applies, so a stored offset is the one actually rendered. Added the rotated, non-square viewport coverage too — a pan → offset → centre round trip plus a drag-off-screen case.

Skipping the `fieldset`/`legend` suggestion. The pattern is indeed used elsewhere in this dialog, but the stated benefit doesn't hold here: a `legend` still occupies a line above the inputs, so the neighbouring "Feature label zoom level" control would still need the alignment spacer. There is also no `fieldset` styling in this component, so the browser default border would box the two fields and change the layout.

joelkoz:
@coderabbitai review

PR #622, opened 2026-07-27
fix(charts): stop the In view filter looping on map view change

### LINKED ISSUE #617: bug: Chart list "in view" locks up FSK when zooming
The "In view" filter added by #352 causes FSK to lock on when active and you zoom in or out.

joelkoz:
Closes #617 — with the chart list's "In view" filter on, panning or zooming the
map locked Freeboard up.

The cause is effect feedback, not the filter logic. Two constructor effects in
`ChartListComponent` both end in `doFilter()`, which writes the `filteredList`
signal and then reads it back via `alignSelections()`:

- `effect(selectedCharts)` → `externalSelection()` → `doFilter()`
- `effect(mapExtent)` → `doFilter()` — added with the filter in #352

Neither effect re-triggers itself, because each reads `filteredList` *after* its
own write, so the version it records is already current. That is why this survived
review. But each one's write invalidates the read the *other* effect recorded, so
once both are live a single `moveend` ping-pongs them without end. Reproduced in a
spec, it runs the JS heap out of memory.

Both effects now run their work inside `untracked()`, so each is triggered only by
the signal it reads outside that call, and the shared list state is a side effect
rather than a dependency. This also stops the `selectedCharts` effect re-running on
every pan and zoom — a second dependency that leaked in via `doFilter()`'s
`mapExtent` read.

Reproducing needs both the `selectedCharts` input bound and the filter on, so the
tests use a real component fixture with the template overridden away; the loop is
in the effects, not the view. The stubbed `arrangeChartLayers` trips after 25 calls
so a regression fails as an assertion instead of running the vitest worker out of
memory.

Verified against a running dev server: the chart list re-filters correctly while
panning and zooming, with no lockup.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where the “In view” chart filter could repeatedly run when the map moved.
  - Improved filtering behavior so map extent changes trigger only the expected updates.
  - Added regression coverage for both enabled and disabled “In view” filtering scenarios.

### DIFF
--- a/src/app/modules/skresources/components/charts/chartlist.spec.ts
+++ b/src/app/modules/skresources/components/charts/chartlist.spec.ts
@@ -121,3 +121,105 @@ describe('ChartListComponent — list ordered by chart layer order (#550)', () =
     expect(idsOf(filteredSignalOf(comp)())).toEqual(['c', 'b', 'a', 'osm']);
   });
 });
+
+/**
+ * Two constructor effects — one on `selectedCharts`, one on `app.mapExtent()` —
+ * both end in `doFilter()`, which writes the `filteredList` signal and then reads
+ * it back (via `alignSelections()`). Tracked, that read made each effect a
+ * dependent of the other's write: with the "In view" filter on, a single map move
+ * ping-ponged them forever and Freeboard locked up on zoom/pan (#617).
+ *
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai review

PR #636, opened 2026-07-28
feat(map): support the cursor bearing/distance/ETA readout on touch devices

### LINKED ISSUE #546: feat(map): support the cursor bearing/distance/ETA readout on touch devices
## What
The status-bar readout — position, zoom, and bearing/distance/ETA from the vessel
to the cursor ("Live ETA at cursor") — is effectively a mouse-only feature. On a
touchscreen it's the device where a quick "how far / what bearing to that point?"
is most valuable, yet it's hard to use there.

## Observed on touch
- The setting is labelled "(mouse only)" and defaults off, so it's hidden from
  touch users to begin with.
- With it enabled, a tap only updates the readout via whatever `pointermove` the
  browser incidentally emits — so taps register inconsistently.
- When the readout does update, there's no marker on the chart, so you can't tell
  which point the bearing/distance refers to.

## Why it matters
Boats are touch-first. This is a genuinely useful "measure to a point" gesture that
today only works well with a mouse.

## Where it lives
- `updateCursorInfo()` — `src/app/modules/map/fb-map.component.ts:715`, called only
  from `onMapPointerMove` (`:658`). `onMapSingleClick` (`:746`) never updates it.
- Readout computation: `src/app/modules/map/cursor-eta.ts`.
- Setting + "(mouse only)" tooltip: `src/app/modules/settings/components/settings-dialog.html:242`;
  default `liveEta: false` in `src/app/app.config.ts`.

## Suggested direction
- Route a tap through the same update path so the readout reliably reflects the
  tapped point on touch.
- Show a brief marker at the tapped point (a dot that fades after a few seconds) so
  the reference location is visible.
- Reconsider the "mouse only" framing / default so touch users can discover it
  (maintainer call — flagging, not assuming).

## Done when
On a touch device, tapping the chart reliably updates the readout and shows a
temporary marker at the tapped point; the setting no longer reads as mouse-only.

Related: route lines hard to select by touch (#548).

---
🤖 Created with [Claude Code](https://claude.com/claude-code)

joelkoz:
The status-bar bearing/distance/ETA readout only ever updated from `pointermove`, which made it effectively mouse-only — the setting even said so. On touch, a tap updated it via whatever pointer events the browser happened to emit, and nothing on the chart showed which point the numbers referred to. That's backwards for a boat: a touchscreen at the helm is where "how far to that?" is most useful.

A short click **or** tap now pins the readout to that point and marks it, so both input types get the same gesture. The pin and the marker expire together, so the numbers are never left on screen with nothing identifying what they describe.

Two gestures are deliberately excluded:

- **Panning** — OpenLayers emits no `singleclick` for a gesture that has dragged, so this needs no guard of its own.
- **Long press** — a duration check, so the chart context menu (and the vertex-delete hold while route editing) still behaves exactly as before.

The marker fades from the moment it appears rather than holding solid first. It is deliberately *not* hit-tested away from features — per the repo's own lessons log, `e.features` is an unreliable guard here because the map's selection tolerance (5px mouse / 15px touch) disagrees with `Modify`'s `pixelTolerance` — so anything it lands on top of is obscured until it clears. Grabbing a route vertex right after a tap was noticeably obstructed at a fixed 3s, which is why the fade is now both immediate and configurable.

**Tap Fade Speed** (Settings → Display → Status Bar Options) sets that duration in seconds; minimum 1, no maximum, default 1.5. The value is validated on read rather than trusted, since app config comes from the server — and clamped to the largest delay `setTimeout` can hold, since past ~24 days it overflows and would fire immediately, inverting the setting's meaning.

The "Live ETA at cursor" setting stays **off by default** — only its "(mouse only)" framing is gone.

### Docs

Both user-facing surfaces are updated here:

- `features/cursor-eta.md` — the existing doc claimed the readout was "a planning aid for anyone at a desktop or laptop rather than something you'll see on a touchscreen at the helm", which this change makes false. Rewritten to current state, with a refreshed screenshot.
- `src/assets/help/index.html` — the Status Bar Options group had never been documented in the settings reference (including the two controls that shipped in v3.0.0). Added at control level, pointing to the Feature Browser for usage rather than restating it.

### Testing

`cursor-marker.spec.ts` covers the pin lifetime (including that it survives the *whole* fade), the per-tap id that restarts the CSS animation, re-tap restarting the countdown, teardown not firing the expiry callback, and fade-duration validation (fallback, clamp-to-minimum, no upper bound, platform cap).

- [x] `npm run test:ci` — 443 passed
- [x] `npm run build:web`
- [x] Verified on a dev server: mouse click, touch tap, pan, press-and-hold, re-tap mid-fade, and the readout holding while lat/lon and zoom keep updating live

Closes #546

## Summary by CodeRabbit

* **New Features**
  * Live ETA at cursor now works on touch devices, allowing users to tap a chart point and temporarily pin its bearing, distance, and ETA readout.
  * Added a configurable **Tap Fade Speed** setting to control how long pinned markers remain visible.
  * Added fallback support using the configured Reference Boat Speed when speed-over-ground data is unavailable.

* **Documentation**
  * Updated feature and help content with touch interaction guidance and new settings.


joelkoz:
@coderabbitai pause

PR #638, opened 2026-07-28
fix(map): hold the follow-vessel offset fixed to the screen, not the boat

mairas:
## Why

#616 made the follow-vessel offset two-axis (ahead + abeam), which is a real
improvement — but it holds the offset in the **vessel's frame**, so it rotates
with the course. The rationale there was that a look-ahead should stay ahead. That
holds underway, but it breaks down exactly when you most often pan to look at
something: **at rest or barely moving, the course is unknown or noisy**, so the
chart swings around the boat as COG jitters (or the offset is suppressed entirely
when there is no course, and follow snaps the vessel back to centre). AvNav and
similar plotters keep the panned offset fixed to the screen for this reason.

Worth noting: in heading-up the two models agree — "up the screen" is along the
course either way — so this only changes behaviour in north-up / low-speed, which
is where the vessel-frame version misbehaves.

## What changes

- The offset is measured in the **screen's frame** (`x` across, `y` up — each a
  percentage of the viewport half-extent) and resolved through the view rotation
  alone, so it never depends on the vessel's course. The boat holds the on-screen
  spot you panned it to whether stationary, drifting, or under way.
- Because the axes are independent screen fractions, a per-axis clamp always keeps
  the vessel on screen, so the corner-scaling machinery (`constrainCenterOffset` /
  `onScreenScale`) is no longer needed and is removed.
- **Panning the vessel off screen releases follow mode** (leaving the chart where
  you left it) rather than clamping it back on.

## Notes

- Config becomes `centerOffset: { x, y }`. The vessel-frame `centerOffset` /
  `centerOffsetAbeam` pair migrates (ahead → y, abeam → x), as do the older
  single-value and 0–0.7 fractional shapes. The two settings fields are now
  **Vertical** / **Horizontal**.
- No version change.

## Testing

- [x] `npm run test:ci` green (full suite); `follow-offset` unit tests cover the
  course-independent screen-frame resolution, the 2D pan round-trip, near-edge
  clamping, off-screen release, and the migration.
- [x] `npm run build:all` clean (AOT template typecheck of the new bindings).
- [x] Interactive: exercised on a live Signal K server — the offset holds its
  on-screen position at rest (the case #616 mishandles) and under way, and follow
  releases when the vessel is panned off screen.

## Summary by CodeRabbit

* **New Features**
  * Map vessel-center offsets now use screen-based vertical and horizontal positioning (x/y percentages).
  * Updated map settings let you configure vertical and horizontal offset percentages instead of ahead/abeam values.
  * While in follow mode, panning now adopts a valid clamped offset or automatically exits follow when the vessel would move out of view.
* **Bug Fixes**
  * Improved migration and coercion of older offset settings to the new x/y format for consistent map positioning.
  * Correct handling of normalized zero values to avoid “negative zero” behavior.
* **Tests**
  * Refreshed unit tests to cover the new offset normalization and pan-follow resolution behavior.

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -1,10 +1,6 @@
 import { FBAppData, IAppConfig, TemperatureUnitDef } from './types';
 import { Convert } from './lib/convert';
-import {
-  clampCenterOffset,
-  legacyCenterOffset,
-  legacyPanBehavior
-} from './lib/follow-offset';
+import { legacyPanBehavior, normaliseCenterOffset } from './lib/follow-offset';
 import { SKVessel } from './modules';
 import { DefaultOptions } from './modules/map/ol/lib/charts/s57.service';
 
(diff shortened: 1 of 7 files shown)


mairas:
Some hooman-written commentary: I tested the #616 version while moored, and after dragging the chart offset, the chart began jumping all over the place because of the COG jitter. This PR replaces the Ahead/Abeam offset concept with an (x, y) offset that is independent of the course. For example, if you're following a coastline, you can set the offset to optimize the marine area visibility, and it doesn't jump around even if the route twists and turns.

Heading-up mode should behave as previously.

I also added a snap-off: if you pan the boat off the screen, follow mode snaps off automatically. This addresses Teppo's concern that you pan to see a completely different location and the chart keeps panning, or the current behavior that prevents panning the boat off-screen.

joelkoz:
@mairas - Can you fix the failed CI? I think they are all code formatting (run `prettier`)

mairas:
> @mairas - Can you fix the failed CI? I think they are all code formatting (run `prettier`)

Whoopsie, will do!

PR #641, opened 2026-07-28
fix(course): convert Arrival Circle field to/from kilometers correctly

### LINKED ISSUE #639: Course Settings "Arrival Circle" field mislabels unit as km when Distance preference is kilometer (value is actually meters)
Version: 3.0.1 (also present on master as of 2026-07)

  **Describe the bug**

  In the Course Settings dialog, the "Arrival Circle" radius field's unit suffix always mirrors the app's Distance unit preference (Settings → Units → Distance), but the underlying
  conversion is only implemented for nautical miles. When the Distance preference is Nautical Miles, the field correctly converts meters ↔ nm for display/entry. When the Distance preference
  is Kilometer, the field skips conversion entirely and displays/accepts the raw meters value — while still labeling the suffix "km".

  Net effect: with Distance set to kilometer, the dialog shows e.g. "100 km" for what is actually a 100 meter arrival circle (the SignalK-spec default for navigation.course.arrivalCircle),
  and typing a new number is stored as meters, not km. This is silently wrong — a user entering "1" intending 1 km actually sets a 1 meter arrival circle.

  Where (for whoever picks this up):

  In the Course Settings modal component (compiled as chunk-RXMZCTQL.js in the 3.0.1 build), ngOnInit:
```
  this.frmArrivalCircle = this.course.courseData().arrivalCircle === null
    ? 0
    : this.course.courseData().arrivalCircle
  this.frmArrivalCircle = this.app.config.units.distance === "kilometer"
    ? this.frmArrivalCircle                                          // <- no m→km conversion
    : Number(We.transform(this.frmArrivalCircle, "m", "naut-mile").toFixed(1))  // m→nm, correct
```
  and the suffix is rendered unconditionally from the preference:
```
  renderSymbol(this.app.config.units.distance)   // "km" whenever preference is kilometer

  The onFormChange save path has the matching bug on the way back to the server:

  let t = this.app.config.units.distance === "kilometer"
    ? Number(e.target.value)                                          // <- treated as meters, not km*1000
    : We.nauticalMilesToKm(Number(e.target.value)) * 1e3               // nm → m, correct
```
  Expected behavior

  When Distance preference is kilometer, the field should convert m ↔ km (value / 1000 for display, value * 1000 on save) to match the label, the same way the nautical-mile branch already
  does for nm.

  To reproduce

  1. Set Settings → Units → Distance to Kilometer.
  2. Open Course Settings.
  3. Note the Arrival Circle field shows "100" with a "km" suffix — but navigation.course.arrivalCircle on the server is 100 (meters), not 100,000 (which 100 km would be).

joelkoz:
## Summary

The Arrival Circle field in Course Settings only converted meters <-> nautical miles for display and save. When the Distance preference was Kilometer, the raw meters value was shown/saved unconverted while still labeled "km" — e.g. a 100 m arrival circle displayed as "100 km", and typing "1" (intending 1 km) set a 1 meter arrival circle.

This also fixes the cached `app.config.course.arrivalCircle` value, which was set from the raw display-unit input instead of the converted meters value — the same value used elsewhere (`activateRoute`, `setDestination`) where it's expected to be in meters.

## Test plan

- [x] Added `course-settings.spec.ts` covering display (m->km, m->nm) and save (km->m, nm->m) conversion; verified it fails on pre-fix code and passes after.
- [x] `npm run test:ci` — all 443 tests pass.
- [x] `npm run build:web` — builds clean.

Closes #639

## Summary by CodeRabbit

* **Bug Fixes**
  * Arrival Circle distances now display correctly according to the selected unit preference.
  * Saved kilometer and nautical mile values are consistently converted to meters.
  * Non-positive distance values continue to be cleared appropriately.

* **Tests**
  * Added coverage for kilometer and nautical mile conversions during display and saving.

### DIFF
--- a/src/app/modules/course/course-settings.spec.ts
+++ b/src/app/modules/course/course-settings.spec.ts
@@ -1,0 +1,85 @@
+import { TestBed } from '@angular/core/testing';
+import { ChangeDetectorRef } from '@angular/core';
+import { beforeEach, describe, it, expect } from 'vitest';
+import { of } from 'rxjs';
+import {
+  MatBottomSheetRef,
+  MAT_BOTTOM_SHEET_DATA
+} from '@angular/material/bottom-sheet';
+
+import { CourseSettingsModal } from './course-settings';
+import { CourseService } from './course.service';
+import { SignalKClient } from 'signalk-client-angular';
+import { AppFacade } from 'src/app/app.facade';
+import { SKStreamFacade } from '../skstream/skstream.facade';
+import { DistanceUnitDef } from 'src/app/types';
+
+/**
+ * Harness for the Arrival Circle field's unit conversion (#639). arrivalCircle
+ * is always stored/streamed in meters (SignalK spec); the dialog must convert
+ * to/from the user's Distance preference on both the display path (ngOnInit)
+ * and the save path (onFormChange).
+ */
+function setup(distanceUnit: DistanceUnitDef, arrivalCircleMeters: number) {
+  const app = {
+    config: {
+      units: { distance: distanceUnit },
+      course: { arrivalCircle: arrivalCircleMeters }
+    },
+    skApiVersion: 2
+  };
+  TestBed.configureTestingModule({
+    providers: [
+      CourseSettingsModal,
+      { provide: AppFacade, useValue: app },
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai pause

PR #646, opened 2026-07-30
fix(map): let touch tap-hold delete a route vertex

mairas:
On touch devices, long-pressing a route vertex in edit mode never deleted it — the vertex only moved.

## Why

A finger long-press makes the browser/WebView emit a **native `contextmenu`** at the OS long-press timeout (~500 ms) — a full second before the 1500 ms vertex-delete hold timer. Its handler (`rightClickHandler`) called `clearTouchTimer()`, so `touchHold` → `tryDeleteHeldVertexOnHold` never ran and the grabbed vertex was only ever dragged. The code assumed OL emits no `contextmenu` for touch — true for OL's synthetic events, but not for the one the WebView raises itself.

## How

While a route is being edited, ignore that native touch `contextmenu` (a small `isTouchContextMenuWhileEditing` predicate) so the hold timer survives. A mouse right-click and a touch long-press when *not* editing are untouched — the not-editing chart menu and measure paths are unchanged. Builds on #644.

The delete currently commits on finger-lift (the existing `vertexDeleteOnRelease` fallback); a follow-up will move it to the hold timeout with haptic + visual feedback.

## Test plan

- [x] `npm run test:ci` green; added unit tests for the predicate (native touch contextmenu while editing → ignored; mouse, not-editing, and the synthetic `touchHold` source → not ignored).
- [x] Verified on an Android WebView (marine plotter): tap-hold now deletes the grabbed vertex.
- [x] Not-editing chart context menu still opens on touch long-press.
- [x] Mouse right-click context menu unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved touch-based vertex editing by filtering native context menus triggered during long-press gestures.
  * Preserved existing right-click behavior for mouse interactions and touch actions outside active editing.
  * Prevented unintended context menus from interrupting vertex deletion gestures.

* **Tests**
  * Added coverage for touch and mouse context-menu scenarios during and outside route editing.

### DIFF
--- a/src/app/modules/map/ol/lib/map.component.ts
+++ b/src/app/modules/map/ol/lib/map.component.ts
@@ -30,6 +30,7 @@ import {
   clearHeldVertexOnRelease,
   clearVertexDeleted,
   clearVertexDeletedOnDrag,
+  isTouchContextMenuWhileEditing,
   startPointerGesture,
   tryDeleteHeldVertexOnHold
 } from './vertex-delete';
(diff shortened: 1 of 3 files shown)


joelkoz:
This was part of #648 (which was merged)

PR #647, opened 2026-07-30
feat(map): show labels on note markers

dgshue:

Note markers never display their name labels, at any zoom.

## How does it work?

Two things prevent it.

`<fb-notes>` is the only resource layer in `fb-map.component.html` not given the
`[labelMinZoom]` / `[mapZoom]` inputs, so the `ngOnChanges` branches that drive
`updateLabels()` in `FBFeatureLayerComponent` never fire.

Binding those alone isn't enough: `ngOnChanges` calls `super.ngOnChanges()` first and
*then* rebuilds every feature from the incoming notes, and unlike the sibling layers
the notes layer's `buildStyle()` doesn't apply the label. Notes are re-fetched as the
map moves, so a label applied by `updateLabels()` is discarded along with the features
it was applied to. This applies it in `buildStyle()`, as `fb-waypoints` does.

## How was this tested?

New spec for the notes layer's labelling contract — the id prefix `updateLabels()`
matches on, the feature name it reads, and the Text placeholder it writes into —
including that a rebuilt feature carries its label without waiting for
`updateLabels()`. That case fails before this change and passes after.

Also verified on a live Signal K server: built from this branch and installed over a
stock 3.0.1 install. Same chart, same view — a tide-station note published by a
plotter extension:

**Before** (stock 3.0.1 — icon renders, no label)

before

**After** (this PR — label renders)

after

Note: `test:ci` on current `master` already fails 8 tests in `whats-new.spec.ts` and
`plotterext.embedding-host.spec.ts`. Unrelated to this change and unaffected by it —
verified by running both specs on a clean `master` checkout.

## Checklist

- [x] One logical change; version numbers untouched.
- [x] Skimmed [`DEV-LESSONS-LEARNED.md`](../docs/freeboard/DEV-LESSONS-LEARNED.md) for traps relevant to this change.
- [x] Tests added/updated for new behaviour and `npm run test:ci` passes.

## Summary by CodeRabbit

* **New Features**
  * Map notes now display labels based on the configured zoom level.
  * Note labels update appropriately as the map zoom changes.
  * Labels appear for notes with icons and for notes using the fallback marker.

* **Bug Fixes**
  * Preserved note labels when map features are rebuilt or refreshed.
  * Improved label visibility when notes do not have an associated icon.

### DIFF
--- a/src/app/modules/map/fb-map.component.html
+++ b/src/app/modules/map/fb-map.component.html
@@ -508,7 +508,13 @@
 
   <!-- Notes (display-filtered by plotter extensions) -->
   @if (showNoteslayer()) {
-    <fb-notes [notes]="plotterExt.visibleNotes()" [zIndex]="450"> </fb-notes>
+    <fb-notes
+      [notes]="plotterExt.visibleNotes()"
+      [labelMinZoom]="this.app.config.map.labelsMinZoom"
+      [mapZoom]="mapZoomLevel()"
+      [zIndex]="450"
+    >
+    </fb-notes>
   }
 
   <!-- Distance & Bearing to -->
--- a/src/app/modules/map/ol/lib/resources/layer-notes.component.spec.ts
+++ b/src/app/modules/map/ol/lib/resources/layer-notes.component.spec.ts
@@ -1,0 +1,103 @@
+import { describe, it, expect } from 'vitest';
+import VectorSource from 'ol/source/Vector';
+import Feature from 'ol/Feature';
+import { Style, Icon } from 'ol/style';
+import { FreeboardNoteLayerComponent } from './layer-notes.component';
+import { FBNotes } from 'src/app/types';
+
+/**
+ * `buildStyle()` applies a note's name as its label when the feature is built,
+ * for both the symbol and fallback-diamond styles. `updateLabels()` — inherited
+ * from FBFeatureLayerComponent and driven by the `labelMinZoom` / `mapZoom`
+ * inputs — then refreshes existing features when the zoom threshold is crossed.
+ *
+ * These are the pieces that have to keep agreeing for a label to appear: the id
+ * prefix `updateLabels()` matches on, the feature name both paths read, and the
+ * Text each style carries. This spec pins that contract so it can't silently
+ * drift apart again.
+ *
+ * Only plain instance fields are touched, so exercise the layer on a bare
+ * prototype instance — no Angular DI needed (same approach as the AIS filter
+ * spec alongside).
+ */
+function layer(fields: {
+  mapZoom: number;
+  labelMinZoom: number;
+  /** omit the symbol to exercise the fallback-diamond style */
+  symbol?: boolean;
+}) {
+  const c = Object.create(
+    FreeboardNoteLayerComponent.prototype
+  ) as FreeboardNoteLayerComponent;
+  Object.assign(c, {
+    source: new VectorSource(),
+    labelPrefixes: ['note'],
+    theme: { labelText: { color: '#000000' } },
+    mapImages: {
+      getSymbol: () =>
+        fields.symbol === false ? undefined : new Icon({ src: 'note.png' })
+    },
+    ...fields
+  });
+  return c as unknown as FreeboardNoteLayerComponent & {
+    source: VectorSource;
+    parseFBNotes: (n: FBNotes) => void;
+    updateLabels: () => void;
+  };
+}
+
+const notes = [
+  [
+    'abc123',
+    {
+      name: 'Cape Fear',
+      position: { latitude: 33.87, longitude: -78.0 },
+      properties: { skIcon: 'anchorage' }
+    }
+  ]
+] as unknown as FBNotes;
+
+function labelOf(source: VectorSource): string {
+  const f = source.getFeatures()[0] as Feature;
+  return ((f.getStyle() as Style).getText()?.getText() as string) ?? '';
+}
+
+describe('notes layer labels', () => {
+  it('labels a note with its name at or above the label zoom threshold', () => {
+    const c = layer({ mapZoom: 12, labelMinZoom: 10 });
+    c.parseFBNotes(notes);
+    c.updateLabels();
+    expect(labelOf(c.source)).toBe('Cape Fear');
+  });
+
+  it('clears the label below the threshold', () => {
+    const c = layer({ mapZoom: 8, labelMinZoom: 10 });
+    c.parseFBNotes(notes);
+    c.updateLabels();
+    expect(labelOf(c.source)).toBe('');
+  });
+
+  it('gives every note feature the id prefix updateLabels() matches on', () => {
+    const c = layer({ mapZoom: 12, labelMinZoom: 10 });
+    c.parseFBNotes(notes);
+    expect(c.source.getFeatures()[0].getId()).toBe('note.abc123');
+  });
+
+  it('labels a note whose icon could not be resolved', () => {
+    // The fallback diamond carries no other identification, so losing the
+    // label there leaves an anonymous marker.
+    const c = layer({ mapZoom: 12, labelMinZoom: 10, symbol: false });
+    c.parseFBNotes(notes);
+    expect(labelOf(c.source)).toBe('Cape Fear');
+  });
+
+  it('labels rebuilt features without waiting for updateLabels()', () => {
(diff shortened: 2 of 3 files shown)


joelkoz:
Thanks for this — and for a really well-put-together first PR. You found both halves of the
bug, which is the hard part: the missing `[labelMinZoom]`/`[mapZoom]` bindings are the obvious
half, but working out that `ngOnChanges` rebuilds the features *after* `super.ngOnChanges()`
has run `updateLabels()` is the part that would have bitten anyone who only fixed the template.
The spec that covers the rebuilt-feature case is exactly the right test to have written, and the
before/after screenshots made the review much easier.

I do want to change the shape of this before it lands, though.

**1. Labels need to be optional.**

Your screenshot has one note in view. Here in the Florida Keys I have dozens in view at normal
navigation zoom — marina notes, bridge clearances, daybeacon notes, plus notes published by
plotter extensions — and with labels always on, they overlap each other and the chart text badly
enough that the chart gets hard to read. `labelsMinZoom` doesn't rescue it, because by the zoom
where labels switch on I'm already in the dense area:

Screenshot 2026-07-31 at 12 59 20 AM

Others may not like this look either, so I'd like a setting to go with this so users with large note groups (ActiveCaptain users for example) can disable it. In Settings → Resources there's already a **NOTES** group
(`id="sectNotes"` in `src/app/modules/settings/components/settings-dialog.html`) holding
"Notes Zoom Level" and "Fetch Notes radius". A dropdown there named **"Note Title"** with:

- `Always Display`
- `Display on Mouse Over`
- `Don't display`

The two `mat-select` blocks already in that group are the idiom to copy — including the
`persistModel()` call pattern. The setting itself goes in `src/app/types/index.d.ts`; the
`display` block is where I'd put it.

For the mouse-over option, `MapComponent` already emits a `pointermove` event (see
`src/app/modules/map/ol/lib/map.component.ts`) — that's the seam. It is mouse-only by nature,
and that's fine: touch users are covered by the existing `singleClickNoteDetails` setting, which
opens the note's details on a single tap.

**2. Put the label below the anchor point.**

Right now it renders on top of the marker — visible in your own "after" screenshot, where
"Fort Caswell" sits across the top of the icon and collides with the tide readout. Worth knowing
the current offset is out of step with the rest of the app too: notes use `offsetY: -12`, while
the equivalent icon branch in `layer-waypoints.component.ts` uses `-28`. Moving the note label
below the anchor instead should clear both the marker and the chart text above it.

**3. One thing to re-check on your test run.**

The note about `test:ci` failing 8 tests on master isn't reproducing for me — I ran
`npm run test:ci` on your base commit and got 462 tests passing, clean. I did reproduce your
failures by running `vitest` directly against `vitest-base.config.ts`, which skips the jsdom
setup and gives `localStorage is not defined`. So I think that's the invocation rather than a red
master. `npm run test:ci` is the one the CI uses — worth re-running to confirm your own checklist
tick.

(For what it's worth, with your branch applied I get 466 passing, and reverting just the
`layer-notes.component.ts` change while keeping your spec does make the rebuilt-features test
fail — so your regression guard genuinely works.)

One process note: CodeRabbit left three findings and none of them have been answered yet. On this
repo I'd like every CR thread disposed of before merge — either fixed, or replied to on the thread
explaining why not. A silent thread doesn't tell me whether you disagreed, missed it, or never
opened it, and where there's a good reason to skip something I'd rather it live in the PR than in
a reviewer's head.

Two of the three are really the same issue: the 🟠 Major on the `RegularShape` fallback, and the
Minor asking for an undefined-icon fixture to cover it. I think CodeRabbit is wrong on that one —
but I'd rather you satisfy yourself than take my word for it. Trace `getSymbol()` in
`map-image-registry.service.ts`: look at what the constructor loads, and at what `getPoiDefs()`
guarantees about the `default` key. If you land where I did, say so on the thread and I'm happy to
leave the code alone. If you don't, then fix it.

The third one — the spec's doc comment still describing `updateLabels()` as the thing that fills
the label, when `buildStyle()` now does — is straightforwardly correct and worth taking.

Adding the setting keeps this within one logical change, so no need to split anything out — it's
still "make note labels work," just with the display control that the feature needs to be usable
on a busy chart. Give me a shout if anything above is unclear.

dgshue:
Thanks — all three addressed in the force-push.

The fallback catch was a real one: `buildStyle()` labelled the symbol branch but returned the RegularShape fallback with no `Text` at all, so a note whose icon failed to resolve rendered as an anonymous diamond. Both branches now go through `setTextLabel()`.

The spec's fixture always returned an `Icon`, which is exactly why it missed that — it now takes a `symbol` flag and covers the unresolved-icon path. Verified red→green: that case fails without the fallback fix and passes with it.

Doc comment corrected to describe `buildStyle()` labelling eagerly at feature creation, with `updateLabels()` refreshing existing features when the zoom threshold is crossed.

`build:all` clean; `test:ci` 459 passing, with the 8 failures in `whats-new.spec.ts` and `plotterext.embedding-host.spec.ts` unchanged from master.

joelkoz:
Thanks — the CodeRabbit items look good, and the fallback branch now matching the symbol branch
is a genuine consistency win.

Two things before this can go further, though. The three CR findings were only part of the round —
the two changes I asked for in my review are still open, and they're the ones blocking merge:
the **"Note Title" setting** and the **label placement**. Without the setting this isn't
mergeable, because on a chart with a lot of notes always-on labels make it unreadable (that was
the screenshot in my last comment).

Also worth flagging: I can't reproduce the failing tests you're reporting. Your branch is
completely green for me — `npm run test:ci` gives **467 passing, 0 failures**, 53 files, on
`40f3ecc5`. I think you're still running `vitest` directly against `vitest-base.config.ts`, which
skips the jsdom setup and produces exactly the `localStorage is not defined` failures you're
seeing in those two specs. `npm run test:ci` is what CI runs — worth switching to before
reporting results.

One correction on the fallback fix, since your description states it as a bug that was happening:
`getSymbol()` can't actually return a falsy icon. `getPoiDefs()` hard-codes the `default` key and
the constructor loads it, so both exit paths always yield an `Icon` and the `RegularShape` branch
is unreachable in practice. Your change is harmless and I'm happy to keep it for consistency —
but no note was ever rendering as an anonymous diamond, and the new spec only reaches that branch
because the fixture forces it. Not asking you to revert anything; just don't carry that away as a
bug that existed.

---

**For your coding agent:**
```
Two changes required. Neither is in the current diff.

**1. Add a "Note Title" display setting.**

- UI: `src/app/modules/settings/components/settings-dialog.html`, in the existing **NOTES**
  group (`id="sectNotes"`, Settings → Resources). Add a `mat-select` labelled `Note Title`.
- Options: `Always Display` / `Display on Mouse Over` / `Don't display`.
- Copy the idiom from the two `mat-select` blocks already in that group — including the
  `persistModel()` call so the choice survives a reload.
- Declare the setting in `src/app/types/index.d.ts` (the `display` block).
- Honour it in `src/app/modules/map/ol/lib/resources/layer-notes.component.ts`. `Don't display`
  must suppress the label regardless of zoom; `Always Display` keeps current behaviour.
- For `Display on Mouse Over`: `MapComponent` already emits a `pointermove` event
  (`src/app/modules/map/ol/lib/map.component.ts`) — that is the seam. It is mouse-only by
  design; touch users are served by the existing `singleClickNoteDetails` setting.

**2. Move the label below the anchor point.**

- Currently `offsetY: -12` in both style branches, which renders the text over the top of the
  marker. It should sit clear below the icon. For reference the equivalent icon branch in
  `layer-waypoints.component.ts` uses `-28` to clear its symbol; pick an offset that clears the
  note icon rather than copying that number.

**Do not** change version numbers, and keep this to the one logical change (note labels +
their display setting) — no unrelated cleanups.

**Definition of done — all of these:**

- [ ] `Note Title` select present in the NOTES settings group, all three options, persisted
      across a reload
- [ ] `Don't display` suppresses note labels entirely; `Always Display` matches current
      behaviour; `Display on Mouse Over` shows the label for the hovered note only
- [ ] Label renders clear of the marker icon, not overlapping it
- [ ] `*.spec.ts` coverage for the new setting's behaviour, including the suppressed case
- [ ] `npm run build:all` clean
- [ ] `npm run test:ci` — **0 failures** (do not run `vitest` against `vitest-base.config.ts`;
      it bypasses the jsdom setup and reports false failures)
- [ ] Every CodeRabbit thread on the PR disposed of — fixed, or answered on the thread with why
      it doesn't apply.
```

PR #652, opened 2026-07-31
fix(resources): prevent resource list effects from looping via alignSelections()

### LINKED ISSUE #623: alignSelections() takes a reactive dependency on the list doFilter() just wrote, inviting the #617 feedback loop
`ResourceListBase.doFilter()` writes the `filteredList` signal and then reads it
back through `alignSelections()`:

```ts
// src/app/modules/skresources/components/resource-list-baseclass.ts
this.filteredList.update(() => fl.slice(0));   // :50
this.alignSelections();                        // :51
...
protected alignSelections() {                  // :57
  this.filteredList().forEach(...)             // :60  <-- tracked read
```

`toggleItem()` does the same thing (`:101` write, `:109` `alignSelections()`).

On its own that is harmless. A single `effect()` that calls `doFilter()` does **not**
loop, because it reads `filteredList` *after* writing it and so records the
already-current version. The failure needs **two** effects that both go through
`doFilter()`/`alignSelections()`: neither re-triggers itself, but each one's write
invalidates the read the other recorded, and they ping-pong until the JS heap is
exhausted. That is #617 — the chart list's "In view" filter locked Freeboard up on
any pan or zoom.

## Why this is worth fixing in the base class

The reactive read serves no purpose. `alignSelections()` is only ever called
immediately after the caller has just set the list; it is recomputing derived
`allSel`/`someSel` flags from state the caller already has. Being a *dependency* of
whatever effect happens to be on the stack is incidental.

Meanwhile the workaround has now been discovered independently twice:

- `routelist.ts:100` wraps `alignCheckedWithCache()` — which writes `filteredList`
  and then calls `alignSelections()` — in `untracked()`.
- `chartlist.ts` does the same for both of its effects in #622, fixing #617.

Seven components extend `ResourceListBase` (charts, routes, waypoints, regions,
tracks, AIS, info layers). Today only the chart list had two effects converging on
`doFilter()`, so only it could ping-pong, but the shape is present in all of them:
the next effect anyone adds to any list re-opens the same trap, and it will present
as a browser lockup rather than as anything pointing at this file.

## Suggested direction

Make the read in `alignSelections()` non-reactive:

```ts
const list = untracked(() => this.filteredList());
```

so subclasses cannot accidentally take a dependency on list state they just wrote,
and the per-caller `untracked()` wrappers become belt-and-braces rather than load
bearing.

Worth confirming while doing it that no subclass actually *wants* `alignSelections()`
to re-run reactively — nothing in the current code does, but it is the one behavioural
question the change turns on. Each list has a spec file to extend.

Split out of #622 (one logical change per PR); that PR fixes #617 at the chart-list
effect level and deliberately does not touch shared code.

joelkoz:
`ResourceListBase.doFilter()` writes the `filteredList` signal and then recomputes
the select all/some/none flags from it via `alignSelections()`. That read back was
**tracked**, so any effect ending in `doFilter()` became a dependent of its own
write.

A single such effect is harmless — it records the version it just produced, so it
never re-triggers itself. Two are not: each one's write invalidates the version the
other recorded, and they ping-pong until the JS heap is exhausted. That is #617,
where the chart list's "In view" filter locked Freeboard up on any pan or zoom.

#622 fixed that at the chart-list effects. Seven components extend
`ResourceListBase`, though, and the shape is present in all of them — the next
effect anyone adds to any list re-opens the trap, and it presents as a browser
lockup with nothing pointing at this file. So the guard belongs in the base class:
`alignSelections()` now reads the list with `untracked()`.

The reactive read served no purpose. `alignSelections()` is only ever called
immediately after the caller has just set the list; being a dependency of whatever
effect happens to be on the stack was incidental. Confirmed no subclass wants it to
re-run reactively — the one override (`AisListComponent`) calls `super` and then
reads `app.config.selections.aisTargetTypes`, a plain array. Fixing it in
`alignSelections()` rather than `doFilter()` matters because `ChartListComponent`
has its own `doFilter()` override that also writes-then-aligns; only the base method
covers every caller.

The existing `untracked()` wrappers in `chartlist.ts` and `routelist.ts` become
belt-and-braces rather than load bearing. They are left in place — removing them
would be a second logical change.

## Test

New `resource-list-baseclass.spec.ts` stands up a minimal `ResourceListBase`
subclass with two effects that both call `doFilter()` **deliberately without** an
`untracked()` wrapper — the mistake the base class has to survive — and asserts the
filter settles. `doFilter()` trips after 25 runs so a regression fails as an
assertion rather than taking the vitest worker down with an OOM crash.

- Fix reverted: 5/5 fail with `doFilter() ran away — alignSelections() feedback loop`
- Fix in place: 5/5 pass
- Full suite: 54 files / 479 tests passed; `build:web` and Prettier clean

Closes #623

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved selection filtering stability to prevent repeated reactive updates and feedback loops.
  - Preserved accurate “select all” and partial-selection states when lists are filtered.

- **Tests**
  - Added comprehensive coverage for initial filtering, repeated triggers, simultaneous updates, filtered results, and selection-state behavior.

### DIFF
--- a/src/app/modules/skresources/components/resource-list-baseclass.spec.ts
+++ b/src/app/modules/skresources/components/resource-list-baseclass.spec.ts
@@ -1,0 +1,134 @@
+import { TestBed } from '@angular/core/testing';
+import { effect, signal } from '@angular/core';
+import { describe, it, expect, beforeEach } from 'vitest';
+
+import { ResourceListBase } from './resource-list-baseclass';
+import type { SKResourceService } from '../resources.service';
+
+/**
+ * `doFilter()` writes the `filteredList` signal and then recomputes the select
+ * all/some/none flags from it via `alignSelections()`. When that read back was
+ * tracked, any effect ending in `doFilter()` became a dependent of its own
+ * write. One such effect is harmless — it records the version it just produced
+ * — but two are not: each one's write invalidates the version the other
+ * recorded, so they re-trigger each other until the heap is exhausted. That was
+ * #617, where the chart list's "In view" filter locked Freeboard up on any pan
+ * or zoom.
+ *
+ * Seven components extend this base class, so the guard belongs here rather
+ * than in each subclass's effects (#623). These tests stand up a minimal
+ * subclass with two effects that both call `doFilter()` *without* an
+ * `untracked()` wrapper — the mistake the base class has to survive — and
+ * assert the filter settles.
+ *
+ * `doFilter()` trips after a small number of calls so a regression fails as an
+ * assertion rather than taking the vitest worker down with an out-of-memory
+ * crash.
+ */
+const RUNAWAY = 25;
+
+type Entry = [string, { name: string }, boolean];
+
+const entry = (id: string, checked: boolean): Entry => [
+  id,
+  { name: id },
+  checked
+];
+
+class TestList extends ResourceListBase {
+  filterRuns = 0;
+  readonly triggerA = signal(0);
+  readonly triggerB = signal(0);
+
+  constructor() {
+    super('routes', {} as SKResourceService);
+    // Deliberately untracked-free: the base class must make these safe.
+    effect(() => {
+      this.triggerA();
+      this.doFilter();
+    });
+    effect(() => {
+      this.triggerB();
+      this.doFilter();
+    });
+  }
+
+  seed(entries: Entry[]) {
+    this.fullList.push(...entries);
+  }
+
+  checkAll() {
+    this.fullList.forEach((e) => (e[2] = true));
+  }
+
+  get list(): Entry[] {
+    return this.filteredList();
+  }
+
+  get flags() {
+    return { all: this.allSel, some: this.someSel };
+  }
+
+  protected override doFilter() {
+    if (++this.filterRuns > RUNAWAY) {
+      throw new Error('doFilter() ran away — alignSelections() feedback loop');
+    }
+    super.doFilter();
+  }
+}
+
+describe('ResourceListBase — alignSelections() takes no reactive dependency (#623)', () => {
+  let comp: TestList;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    comp = TestBed.runInInjectionContext(() => new TestList());
+    comp.seed([entry('a', true), entry('b', false)]);
+  });
+
+  it('settles after both effects have run once', () => {
+    TestBed.tick();
+
+    expect(comp.filterRuns).toBe(2);
+  });
+
+  it('re-filters once when a single effect is triggered', () => {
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai pause

joelkoz:
Re: the nitpick on `resource-list-baseclass.ts#L56-L62` — **applied** in be7382a8.

You were right that the incident narrative did not belong in a source comment; the guideline about describing current state rather than development history covers it. Both comments now state the contract instead:

- The base-class comment says a tracked read would make any effect on the call stack a dependent of the `filteredList` write, that two such effects re-trigger each other until the heap is exhausted, and that `untracked()` is therefore load bearing.
- The spec header says the two effects verify that filtering settles and the selection flags stay correct.

One thing deliberately kept: the bare `(#617)` pointer. The reason is not history but discoverability — this is a guard whose removal fails as a browser lockup with nothing in the stack pointing at this file, so a reader deciding whether `untracked()` is safe to simplify needs one hop to the evidence. A single issue number is the cheapest way to leave that without narrating the incident inline.

joelkoz:
@coderabbitai — adding a general rebuttal to record alongside the above, since this will come up again on this project.

**The "describe current state, not development history" guideline needs an explicit exception for defensive code.**

Most code exists to *do* something, and for that code the guideline is right: how it came to look the way it does is noise, and the git history holds it. But a defensive guard is different in kind. It exists solely to *prevent* something, so:

- **Its rationale is its contract, not its history.** `untracked()` here produces no behaviour a reader can observe. Every test still passes if you remove it. The only thing that tells a future developer it is load bearing is the note saying what breaks without it — that note is a statement of the current contract, in the most literal sense.
- **Stripping the rationale is what causes the regression.** Code whose purpose is invisible reads as redundant, and redundant-looking code gets "simplified" in good faith during an unrelated cleanup. A guard documented only as *what it does* (`reads the list non-reactively`) invites exactly that. Documented as *what it prevents*, it does not.
- **The failure mode determines how much pointer you need.** Where removing a guard fails loudly — a red test, a type error — the code is self-policing and a bare comment is enough. This one fails as a browser lockup with an exhausted heap and nothing in the stack naming this file. A single issue reference is the cheapest possible way to leave the evidence one hop away, and it is not a substitute for the explanation, which stays inline.

So the test we would ask you to apply is not *"does this text mention the past?"* but *"would a reader who deleted this code know they had broken something?"* When the answer is no, the note explaining what it prevents — including a bare issue reference as evidence — is current-state documentation and should not be flagged as history. Narrating the incident inline (what the user saw, which release, how it was found) is still out of scope, and that part of your finding was correct and has been applied.

PR #656, opened 2026-07-31
fix(plotterext): make the no-settings dialog message legible in dark mode

### LINKED ISSUE #653: bug: Plotter-extension remove dialog: "This widget has no settings" text invisible in dark mode
### Summary

The plotter-extension **remove dialog** for a widget that declares no `configPanel` shows the message **"This widget has no settings."** in text whose color is hard-coded to a near-black value. On the dark-themed dialog in dark mode the text is invisible — only the red "Remove widget" button is visible, so the dialog looks empty apart from the button.

### Version

v3.1.0-beta.2

### Steps to reproduce

1. Run Freeboard-SK in **dark mode**.
2. Install any plotter-extension that contributes a widget with **no `configPanel`** (e.g. the Wind Steer widget from the `@halos-org/skip` extension), and place the widget on the chart.
3. Long-press the widget to open its dialog (the remove-only dialog, since the widget has no config panel).

**Observed:** the "This widget has no settings." line is not visible (dark text on dark surface). **Expected:** the message is legible in both light and dark themes.

### Cause

`src/app/modules/plotterext/panel-dialog.component.ts` — the `@else` (no-config) branch renders the message inside `.pe-panel-noconfig`, whose style hard-codes a light-theme text color:

```
.pe-panel-noconfig {
  padding: 16px 24px;
  color: rgba(0, 0, 0, 0.6);   // <-- not theme-aware; invisible in dark mode
}
```

The color is a fixed black; nothing switches it for the dark theme.

### Suggested fix

Use a theme-aware color instead of the hard-coded `rgba(0, 0, 0, 0.6)` — e.g. the Material M3 supporting-text token `var(--mat-sys-on-surface-variant)`, or simply drop the explicit `color` so the text inherits the dialog's themed foreground. (For consistency, worth a glance at `.pe-remove`'s `#d32f2f` too — it happens to read on both themes, but an `error` token would be more robust.)

joelkoz:
The "This widget has no settings." message pinned `color: rgba(0, 0, 0, 0.6)`, which is a copy of Angular Material's *light-mode fallback* for dialog supporting text:

```css
.mat-mdc-dialog-content {
  color: var(--mat-dialog-supporting-text-color, var(--mat-sys-on-surface-variant, rgba(0, 0, 0, 0.6)));
}
```

Pinning that value overrode `--mat-dialog-supporting-text-color`, which `all-component-colors($dark-theme)` *does* emit under `.dark-theme`. The dialog chrome therefore went dark while the message stayed near-black, leaving the dialog looking empty apart from the red Remove button.

The element is already a `mat-dialog-content`, so declaring no colour of its own is the whole fix — the themed supporting-text colour then applies in both themes, and that colour is `on-surface-variant`, the same muted look the hard-coded value was reaching for.

Deliberately *not* using `var(--mat-sys-on-surface-variant)` as the issue suggests: FSK's `define-theme` + `all-component-themes` setup never emits the Material 3 system palette, so `--mat-sys-*` tokens always resolve to their light fallback in both themes — see *Component surfaces styled from `--mat-sys-*` tokens go light in dark mode* in the lessons log.

`.pe-remove`'s `#d32f2f` is left as-is: it reads on both themes, and it is a separate concern from the invisible text.

Verified by the new spec, which fails before the fix (`expected 'rgba(0, 0, 0, 0.6)' to be 'rgb(1, 2, 3)'`) and passes after. It sets a colour on an ancestor to stand in for the dialog's themed foreground, so any future re-pinning of a colour on this element trips it.

Closes #653

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the no-configuration dialog message to inherit the active theme’s text color, ensuring proper readability in dark mode.

* **Tests**
  * Added coverage verifying the message content and theme-aware text styling.

### DIFF
--- a/src/app/modules/plotterext/panel-dialog.component.spec.ts
+++ b/src/app/modules/plotterext/panel-dialog.component.spec.ts
@@ -1,0 +1,60 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import { PlotterPanelDialog } from './panel-dialog.component';
+import { PlotterExtensionService } from './plotterext.service';
+
+describe('PlotterPanelDialog with no configuration panel', () => {
+  let fixture: ComponentFixture<PlotterPanelDialog>;
+
+  const message = () =>
+    fixture.nativeElement.querySelector('.pe-panel-noconfig') as HTMLElement;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [PlotterPanelDialog],
+      providers: [
+        provideNoopAnimations(),
+        { provide: MatDialogRef, useValue: { close: () => undefined } },
+        {
+          provide: MAT_DIALOG_DATA,
+          useValue: {
+            extension: 'test-ext',
+            panel: null,
+            title: 'Wind Steer',
+            targetInstance: 'inst-1',
+            targetWidget: 'wind-steer'
+          }
+        },
+        {
+          provide: PlotterExtensionService,
+          useValue: {
+            resolveAssetUrl: (u: string) => u,
+            attachPanel: () => () => undefined,
+            removeWidget: () => undefined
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai pause

PR #663, opened 2026-08-01
fix(skresources): let the symbol pickers scroll on touch

### LINKED ISSUE #651: bug: Note symbol selection dropdown can't be scrolled on touch devices
## Summary

When adding a new note, the **symbol selection dropdown cannot be scrolled on a touch interface**. The list of symbols opens, but touch-dragging inside the dropdown does not scroll it, so symbols below the initially visible ones are unreachable. This is worse with **"Show all symbols"** enabled, where the list is long enough that most symbols are off-screen.

Freeboard runs on marine touchscreen chartplotters and tablets, where this is often the only input method, so the symbol picker is effectively unusable there.

## Steps to reproduce

1. On a touch device (tablet, phone, or touchscreen display), open Freeboard-SK.
2. Add a new note.
3. In the note dialog, tap the symbol/icon selector to open its dropdown.
4. Optionally enable **"Show all symbols"** to get the full list.
5. Try to scroll the dropdown by dragging with a finger.

## Expected

The symbol dropdown scrolls with touch, exposing all symbols in the list.

## Actual

The dropdown does not scroll under touch; only the symbols visible when it opens can be selected.

## Notes

The selector is the Angular Material `mat-select` in the note dialog
(`src/app/modules/skresources/components/notes/note-dialog.html`, the `<mat-select>` bound to `icon.svgIcon` / `poiIcons`). Its options render in a CDK overlay panel with a capped `max-height` and internal scroll region — a common spot for touch `touchmove` scrolling to fail inside the overlay. Reproduces most reliably once the option count exceeds the panel height (i.e. with "Show all symbols" on).

joelkoz:
Freeboard's symbol pickers could not be scrolled with a finger, so every symbol
below the fold was unreachable — worst with "Show all symbols" on. On the tablets
and touchscreen chartplotters much of our audience steers with, that is the only
input method, which left the pickers effectively unusable.

`MatTooltip` claims the long press on a touch platform by writing an inline
`touch-action: none` onto its trigger element. `touch-action` is read from
wherever a touch *begins*, so a trigger occupying every row of a scrolling list
leaves a finger nowhere to land that pans it. Both pickers are built exactly that
way — every option of the note dialog's dropdown and every tile of the waypoint
dialog's grid carries a tooltip — so neither scrolled. The desktop `mouseenter`
branch never writes that style, which is why this never appeared with a mouse.

Turning the tooltip's touch gestures off restores the scroll. The cost is the
long-press symbol name on touch, which could not be used anyway while it was the
thing blocking the scroll.

The waypoint picker is included because it is the same defect in the same
feature, one attribute away, and would be hit immediately after the note dialog
alone was fixed.

The test renders both dialogs with the platform reported as iOS — the condition
that selects the touch branch inside the directive — and asserts no picker row
carries `touch-action: none`. Without the fix it flags all 21 options and all 20
tiles.

No screenshots: nothing changes visually, only whether a drag scrolls. Verified
on a physical touch device — both pickers scroll.

Note on CI: the advisory armv7 (Cerbo GX) leg fails on the 900s `test:ci` hard
timeout. That cap is already nearly exhausted on `master` (763s measured at 53
spec files), so it is tripped by the addition of any spec file rather than by
anything specific to this one — the suite is unchanged in cost on every other
platform. Filed separately.

Closes #651

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved touch scrolling in note and waypoint icon pickers.
  * Prevented icon tooltips from interfering with scrolling on touch devices.
  * Retained desktop tooltip labels and icon selection behavior.

* **Tests**
  * Added regression coverage for touch scrolling in both symbol pickers.

### DIFF
--- a/src/app/modules/skresources/components/notes/note-dialog.html
+++ b/src/app/modules/skresources/components/notes/note-dialog.html
@@ -253,7 +253,14 @@
                   }
                 </mat-select-trigger>
                 @for(i of poiIcons; track i.id) {
-                <mat-option [value]="i.id" [matTooltip]="i.name"
+                <!-- Touch gestures off: on a touch platform MatTooltip sets
+                inline touch-action:none on its trigger to reserve the long
+                press, and every row of this panel is a trigger — so a finger
+                has nowhere to land that scrolls the list (#651). -->
+                <mat-option
+                  [value]="i.id"
+                  [matTooltip]="i.name"
+                  matTooltipTouchGestures="off"
                   ><mat-icon [svgIcon]="i.id">{{i.name}}</mat-icon>
                 </mat-option>
                 }
--- a/src/app/modules/skresources/components/symbol-picker-touch.spec.ts
+++ b/src/app/modules/skresources/components/symbol-picker-touch.spec.ts
@@ -19,12 +19,13 @@ import { WaypointDialog } from './waypoints/waypoint-dialog';
  * On a touch platform `MatTooltip` reserves the long press for itself by
  * writing an inline `touch-action: none` onto its trigger element
  * (`_disableNativeGesturesIfNecessary`). `touch-action` is read from wherever a
- * touch *starts*, so a trigger that fills every row of a scrolling list leaves
- * the finger nowhere to land that pans it: the note dialog's symbol dropdown
- * and the waypoint dialog's symbol grid both became unscrollable, stranding
- * every symbol below the fold. The desktop path takes the `mouseenter` branch
- * and never writes the style, which is why this only shows up on a tablet or a
- * chartplotter — for many Freeboard users, their only input device.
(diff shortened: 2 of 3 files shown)


joelkoz:
The armv7 CI note in the description is filed as #664.

joelkoz:
@coderabbitai pause

joelkoz:
Nitpick on `symbol-picker-touch.spec.ts` (lines 16-31) — **fixed** in 46088b25.

Fair catch against the repo's "documentation describes current state, not development history" rule. The docblock now states what the pickers are and what they do about it — every row of both carries a tooltip, so each turns touch gestures off — rather than narrating that they "became unscrollable". The explanation of why the tests report the platform as iOS is unchanged.

Reviews are paused for this push since it only touches that comment; no code changed.

PR #678, opened 2026-08-07
fix(charts): make the adjustment palette usable beside the chart list

mairas:

Opening the Image Adjustment palette closed the chart list, so adjusting several
charts in a stack meant reopening the list every time.

Making the palette modeless exposed two problems the old behaviour had been
hiding. The saved drag position is relative to where the palette opens, and that
origin moved — an offset saved against the old one could put the palette's drag
handle, the only way to move it back, past the viewport edge. And with the list
reachable, a second palette could open on top of the first; closing the covered
one reverts an adjustment the visible one already saved.

Two independent bugs fixed while in here: the palette's text was unreadable in
dark mode (Angular Material themes the dialog surface but not its colour, so the
palette's own markup inherited black on a near-black surface —
`rgb(0,0,0)` on `rgb(24,18,17)`), and a long chart name wrapped to two lines and
collided with the line below it.

## How does it work?

The palette opens beside the chart list, which stays put. Saved positions are
clamped back on screen on both axes, and only one per-chart palette is open at a
time. The dialog takes `--mat-app-text-color`, which follows the theme — measured
`rgb(237,224,221)` dark, `rgb(26,27,31)` light. The header is single-line with an
ellipsis and the full name on hover.

## What changes for the user?

Adjusting the brightness and contrast of several charts is now one pass instead of
one round trip per chart. Open **Charts**, press the **Image Adjustment** button
(the sliders icon) on a row, and the palette appears beside the list rather than
replacing it — so the next chart's button is still there, one click away. Pressing
it on a second chart moves the palette to that chart instead of stacking a second
one on top.

The palette can be dragged anywhere and remembers where it was put. If it was last
left near the edge of a larger window, it now comes back on screen rather than
somewhere its title bar cannot be reached. In dark mode its text is legible; it
used to be black on near-black. A chart with a long name no longer pushes the
palette's own controls out of line.

## How was this tested?

`resources-open-image-adjustment.spec.ts` covers the clamping at wide and narrow
viewports. The existing "opens at the remembered position" test passed only
because jsdom's default width left the saved offset in range, so it no longer
proved anything on its own.

CI green across the matrix; `npm run build:web` and `npm run test:ci` clean
locally too, bar two specs that fail the same way on an untouched `origin/master`
here and pass in CI.

## Checklist

- [x] One logical change; version numbers untouched.
- [x] `features/` untouched — the user-facing change is described above instead.
- [x] Skimmed [`DEV-LESSONS-LEARNED.md`](../docs/freeboard/DEV-LESSONS-LEARNED.md) for traps relevant to this change.
- [x] Tests added/updated for new behaviour and `npm run test:ci` passes.

---

**First of three.** #679 and #680 are stacked on this branch; merge order is
**#678 → #679 → #680**. Each stands on its own.

## Summary by CodeRabbit

* **New Features**
  * Image-adjustment palettes remain open alongside the chart list for easier editing.
  * Palettes open beside the chart list and stay within the visible screen area.
  * Opening a new palette automatically closes the previously open one.
  * SAVE keeps image adjustments, while cancellation restores the original values.
* **Improvements**
  * Long adjustment titles display on one line with an ellipsis; hovering reveals the full title.
  * Improved text visibility within the image-adjustment dialog.
  * Palette replacement now behaves reliably when switching between charts.

### DIFF
--- a/src/app/lib/components/dialogs/image-adjustment-dialog.ts
+++ b/src/app/lib/components/dialogs/image-adjustment-dialog.ts
@@ -40,7 +40,7 @@ const toRatio = (percent: number) => percent / 100;
   template: `
     <div
       class="_ap-image-adjustment"
-      style="overflow: hidden;"
+      style="overflow: hidden; color: var(--mat-app-text-color);"
       cdkDrag
       cdkDragRootElement=".cdk-overlay-pane"
       [cdkDragFreeDragPosition]="initialPosition"
@@ -51,7 +51,18 @@ const toRatio = (percent: number) => percent / 100;
         style="display:flex; align-items:center; cursor:move; padding: 4px 4px 0 12px;"
       >
         <mat-icon style="opacity:0.6;">tune</mat-icon>
-        <div style="flex: 1 1 auto; padding-left: 8px; font-weight: 500;">
+        <div
+          style="
+            flex: 1 1 auto;
+            min-width: 0;
+            padding-left: 8px;
+            font-weight: 500;
+            overflow: hidden;
+            text-overflow: ellipsis;
+            white-space: nowrap;
+          "
+          [title]="data.text"
+        >
           {{ data.text }}
         </div>
         <button mat-icon-button aria-label="Close" (click)="handleClose(false)">
--- a/src/app/modules/skresources/components/charts/chartlist.ts
+++ b/src/app/modules/skresources/components/charts/chartlist.ts
@@ -317,10 +317,9 @@ export class ChartListComponent extends ResourceListBase {
   }
 
   protected itemImageAdjustment(chart: FBChart) {
-    // The palette is modeless and owned by the service so it survives the chart
-    // list closing, which frees the map for the live adjustment preview.
+    // The palette is modeless and owned by the service, and the list stays open
+    // beside it: adjusting one chart of a stack usually means adjusting more.
     this.skres.openImageAdjustment(chart);
-    this.close();
   }
 
   updateFullList(chart: FBChart) {
--- a/src/app/modules/skresources/resources-open-image-adjustment.spec.ts
+++ b/src/app/modules/skresources/resources-open-image-adjustment.spec.ts
@@ -123,32 +123,46 @@ describe('openImageAdjustment (#457)', () => {
     // The palette now opens clear of the chart list, so an offset saved against
     // the old origin can put the drag handle -- the only way to move it back --
     // past the viewport edge.
-    const withViewport = (width: number, fn: () => void) => {
-      const original = window.innerWidth;
-      Object.defineProperty(window, 'innerWidth', {
-        value: width,
-        configurable: true
-      });
+    const withViewport = (
+      width: number,
+      height: number | undefined,
+      fn: () => void
+    ) => {
+      const original = {
+        width: window.innerWidth,
+        height: window.innerHeight
+      };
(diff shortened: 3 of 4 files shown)


joelkoz:
@coderabbitai review

mairas:
Rebased onto current `master` and restructured so this PR stands on its own, per
the landing plan on #680.

**`ae249d81` moved here** (now `f80d7ad2`) — the palette-position clamp belongs
with the palette change that exposed it, not three PRs downstream.

**`features/` reverted.** #682 landed while I was working, so this now follows the
rule rather than just your comment: no feature doc, no `changelog.json` row. The
user-facing description stays in the PR body above.

That leaves two commits: the palette itself, and the clamp.

Verified: `npm run build:web` clean, `npm run test:ci` 8 failures — all of them
`whats-new.spec.ts` and `plotterext.embedding-host.spec.ts`, which fail
identically on a clean `origin/master` checkout. Nothing in this stack touches
either.

mairas:
Typing fix landed here rather than downstream, as asked — `a86312e1`.

The two suppressions were carrying more than a lint waiver. `openChartPalette()`
took the component as `any`, so nothing checked it was openable, and the result
type came back untyped: `afterClosed()` handed out an
`ImageAdjustmentDialogResult` on the call site's say-so alone. The component is
now a `ComponentType<T>` and the result a second type parameter, so the call
site names both and gets a checked `MatDialogRef<T, R>` back.

The stored handle stays `MatDialogRef<unknown, unknown>` rather than the union
CodeRabbit suggested. It is only closed and compared by identity, never read
through, so `unknown` says exactly what the code relies on — and a union would
have to name every palette component, which is a list this field does not need
to know.

`data` is `MatDialogConfig<unknown>`: each palette declares its own shape via
`MAT_DIALOG_DATA`, and the service has no reason to restate it. (On #679 that
shape gets a name, `ChartMinZoomDialogData`, so the dialog and its test fixture
stop keeping separate copies of it.)

**Verification.** `npm run build:web` clean. `npm run test:ci`: 501 passing, 0
failing.

mairas:
The palette stale-close finding from the latest review lands here — `e8a11ddf`
— since the palette mechanism is this PR's.

**It is real, and the reachability argument is the hardware rather than human
speed.** `afterClosed()` arrives an exit animation after the replacement is on
screen; the ordering is the one `releaseChartPalette()` already documents. In
that window the old palette's cancel path put the pre-edit adjustment back
unconditionally, over whatever the replacement had previewed — and if the
replacement was applied inside the window, the config kept the new value while
the map showed the old one until the next list refresh. No human beats a dialog
animation on a desktop, but on a Pi driving its own browser the frames stretch
and a queued second click lands well inside it.

CodeRabbit's warning about the fix was the useful part: keying it by chart
alone would be wrong, because a palette displaced by one for *another* chart
must still revert its own. Palettes now carry what they edit
(`<setting>:<chart id>`), so only a newer palette over the same edit suppresses
the revert. That distinction is what a per-chart session would have lost, and
it is why #679's minimum-zoom palette taking over the same chart still lets the
adjustment revert — the tests for that pair live on #679, where both palettes
exist.

Two tests here: a replacement over the same chart leaves the preview alone, and
a replacement for another chart still reverts. The first fails against the
previous behaviour.

**Verification.** `format:check` clean, `build:web` clean, `test:ci` 503
passing, 0 failing.

PR #683, opened 2026-08-08
fix(instruments): allow form submission in the instrument panel iframe

### LINKED ISSUE #671: "kip" running as the instruments plugin can't login in chrome due to iframe permissions
i run kip as the "instruments" plugin in freeboard-sk. in this configuration with the panel expanded it seems impossible to login to kip recently.

it seems like kip runs in an iframe in fsk and the iframe doesn't have the correct permissions to display a form?

this is in google chrome, i think it works okay in firefox. 

Image

joelkoz:
KIP running as the instrument panel can't be signed in to under Chrome — the sign-in form submits, nothing happens, and no error surfaces.

The instrument panel iframe carried `sandbox="allow-scripts allow-same-origin"`, so Chrome blocks form submission from it. The reporter's console shows `Blocked form submission ... the 'allow-forms' permission is not set`, sourced to `/@mxtommy/kip/#/options`. It isn't KIP-specific: any webapp set as an instrument panel favourite that uses a real form hits the same wall.

The plotter extension iframes (widget, panel drawer, panel dialog, background runtime) already use `allow-scripts allow-same-origin allow-forms`. This brings the instrument panel — the one remaining iframe in the app — up to that same baseline.

No security tradeoff: `allow-same-origin` together with `allow-scripts` already makes the sandbox fault containment rather than a boundary, since such a frame can clear its own sandbox attribute via the parent. Navigation, popups and modals stay withheld.

Verified by signing in to KIP through the instrument panel in Chrome against a local server.

The added spec asserts the exact sandbox token set, so it fails both if `allow-forms` goes missing and if a later change grants something broader. jsdom neither enforces sandbox nor reflects it onto the `sandbox` DOMTokenList, so the assertion reads the attribute — the browser-level behaviour itself isn't reachable from the unit suite.

Closes #671

## Summary by CodeRabbit

* **New Features**
  * Instrument panel forms can now be submitted directly within the embedded panel.

* **Tests**
  * Added coverage to verify the embedded panel uses the expected sandbox permissions.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -409,10 +409,13 @@
                 "
               >
                 @if (app.instrumentPanel().activate) {
+                  <!-- allow-forms: embedded webapps submit forms (e.g. KIP's
+                       Signal K sign-in). Matches the plotter extension iframe
+                       baseline — fault containment, not a security boundary. -->
                   <iframe
                     [src]="instUrl()"
                     style="width: 345px; height: 100%"
-                    sandbox="allow-scripts allow-same-origin"
+                    sandbox="allow-scripts allow-same-origin allow-forms"
                   >
                   </iframe>
                 }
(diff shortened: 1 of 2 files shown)


joelkoz:
@coderabbitai review

PR #685, opened 2026-08-08
chore(lint): replace unsupported eslintrc config with flat config

### LINKED ISSUE #684: ESLint config is eslintrc-era while ESLint is pinned to v10, so every lint run fails
The repository still carries eslintrc-era configuration (`.eslintrc` + `.eslintignore`) while `package.json` pins `"eslint": "^10.5.0"`. ESLint 9 made flat config the default and ESLint 10 removed eslintrc support entirely, so **every** ESLint invocation against this repo fails before linting anything.

### Reproduction

```bash
npx eslint src/app/app.component.spec.ts
```

```
(node:84033) ESLintIgnoreWarning: The ".eslintignore" file is no longer supported.
ESLint couldn't find an eslint.config.(js|mjs|cjs) file.
```

And when a tool discovers `.eslintrc` and points ESLint at it explicitly:

```bash
npx eslint -c .eslintrc src/app/app.component.spec.ts
```

```
.eslintrc:2
    "root": true,
          ^
SyntaxError: Unexpected token ':'
    at wrapSafe (node:internal/modules/cjs/loader:1713:18)
    at Module._compile (node:internal/modules/cjs/loader:1755:20)
```

`.eslintrc` has no file extension, so Node's CommonJS loader parses it as JavaScript: `{` opens a block, `"root"` is a string expression, and the `:` is a syntax error. It never gets far enough to notice the file is JSON.

### Impact

CodeRabbit runs ESLint as part of its review and hits the second failure on every PR, reporting it under "issues while running some tools". That static-analysis input is therefore missing from every review the project receives — a silent degradation that reads as "clean" rather than "not checked".

### Why it went unnoticed

There is no `lint` script in `package.json` and no lint step in any workflow, so ESLint has no consumer inside the repo. Dependabot has bumped it 8 → 9 → 10, straight through both breaking releases, with nothing to signal the config had stopped working.

### Additional staleness

- `.eslintignore` lists `dist` and `e2e`; neither exists. The actual build outputs are `public/` and `plugin/`.
- `.eslintrc` contains a `//` comment inside JSON.
- The config predates the current codebase — it has two commits, the older one titled `v2`.

### Proposed fix

- Add `eslint.config.js` in flat format. `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin` (v8) are already devDependencies and ship flat-config support.
- Delete `.eslintrc` and `.eslintignore`, moving ignores into `ignores` (`public`, `plugin`, `node_modules`).
- Add a `lint` script so the configuration is actually exercised and cannot rot invisibly again.

### Open question — scope

Restoring a working ESLint will almost certainly surface pre-existing violations across `src/`. Landing the config and fixing every violation in one change would be large and would mix concerns. Preference is to land the working configuration first (and, if the volume warrants, keep it non-blocking initially), then address violations separately.

joelkoz:
`eslint` is pinned to `^10.5.0`, but the repo still carried `.eslintrc` and `.eslintignore`. ESLint 9 made flat config the default and v10 removed eslintrc support, so every invocation failed before linting anything — pointed at the extensionless `.eslintrc`, Node's CommonJS loader parses it as JavaScript and dies on the colon in line 2.

Nothing in the repo runs ESLint (no `lint` script, no CI step), so this stayed invisible while the dependency was bumped 8 → 9 → 10 through both breaking releases. CodeRabbit does run it, and has been reporting the failure on every PR — which quietly removed static analysis from every review the project received.

Replaces the legacy files with `eslint.config.js` mirroring the previous rule selection (`eslint:recommended` + `@typescript-eslint` recommended, `eqeqeq` downgraded to a warning), and adds a `lint` script so the config has a consumer and can't rot unnoticed again. The old ignore list named `dist` and `e2e`, neither of which exists; the flat config ignores the real build outputs (`public`, `plugin`).

`@eslint/js` is added as a devDependency — it is the flat-config equivalent of `eslint:recommended` and isn't resolvable on its own.

**Not wired into CI.** With ESLint working again, `npm run lint` reports 413 pre-existing violations (341 errors, 72 warnings) across 104 files, all in `src/app` apart from two in `helper/`. Fixing those is separate work; gating CI on it belongs with that change. No `src/` files are touched here.

Closes #684

@coderabbitai ignore


joelkoz:
Follow-up for the violations this exposes: #686

PR #697, opened 2026-08-08
fix(stream): alarm on a detected disconnect instead of disarming the watchdog

### LINKED ISSUE #695: Connection watchdog is disarmed by a clean disconnect, so the "No server messages!" indicator never appears
The status bar carries a red "No server messages!" icon driven by the stream worker's `watchDog`, which alarms after 18 empty 500ms message intervals (~9s). It works when the link dies *silently* — but it is disarmed in exactly the case where the server is known to be gone.

## What happens

On a detected close, `handleStreamEvent`'s `onClose` branch does two things, and both prevent the alarm:

```ts
case 'onClose':
  closeStream(false);        // calls clearTimers() — stops the interval that raises the alarm
  watchDog.active = false;   // and the alarm is gated on `watchDog.active`
  break;
```

(`src/app/modules/skstream/skstream.worker.ts` — `onClose` around L190, `closeStream`'s `clearTimers()` at L356, the alarm gate at L742.)

So if the websocket closes cleanly and promptly — a server shutdown or restart sending a proper close frame — the alarm never raises, because the ~9s of silence needed to trip it can never elapse. The indicator appears only when the socket stays open with nothing arriving, which is the *harder* case to detect, not the obvious one. `onError` (L203) disarms the same way.

The stale-position indication added in #693 does not share this flaw — it measures position age from a receipt stamp and is unaffected by socket state — so the boat position on the chart is now flagged either way. This issue is specifically about the status-bar server-connection indicator being silent on a clean disconnect.

## Expected

A clean, detected close should be an immediate "no server messages" condition, not a reason to stop watching for one. Setting the alarm on `onClose` rather than clearing `active` would be one way; the reconnect path already resets `alarm = false` / `active = true` on `onConnect` (L184–187), so recovery is already handled.

## Reproduce

1. Connect Freeboard to a Signal K server.
2. Stop the server cleanly (so the browser receives a close frame rather than a dropped link).
3. The red status-bar icon does not appear.

Compare with pulling the device's wifi, where no close frame arrives, the timers keep running, and the icon does appear after ~9s.

Found while working on #672 / #693.

joelkoz:
The status bar's red "No server messages!" indicator was unreachable in exactly the case where the server is known to be gone — a clean shutdown or restart that sends a close frame. It appeared only for a socket that stayed open with nothing arriving, which is the harder case to detect, not the obvious one.

`onClose` ran `closeStream()`, which clears the interval that measures the silence, and then set the watchdog inactive — so the ~9s of quiet needed to trip the alarm could never elapse. `onError` disarmed the same way, and there the timers keep running, so the interval's reset branch would have cleared any alarm set on that path anyway.

A detected close or error *is* the no-messages condition the watchdog exists to report, so both branches now raise the alarm and post it to the app immediately, leaving the watchdog armed. `onConnect` already clears the alarm and re-arms on reconnect, so recovery is unchanged. A deliberate restart is unaffected: `cmd: 'close'` unsubscribes from the stream events before closing the socket, so it never re-enters `handleStreamEvent`.

The online help described the old trigger — "messages ... have been received over the open connection in the last 8-10 seconds" — which this change makes wrong, since the indicator now also appears immediately on a close. That passage is corrected here rather than left to a later docs pass (and a typo in the same sentence goes with it).

Regression tests cover both paths, plus a guard that the `close` message the app reconnects on is still posted. `handleStreamEvent` and `initVessels` are exported as test seams, following the existing pattern in this worker.

Closes #695

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection monitoring so alarms are raised when the server connection closes or encounters an error.
  - Preserved the reconnection message when a connection closes.
  - Ensured monitoring remains active and updates immediately when an alarm is raised.

- **Documentation**
  - Clarified the “No server messages” help text, including connection closures, inactivity lasting 8–10 seconds, and indicator clearing after reconnection.

### DIFF
--- a/src/app/modules/skstream/skstream.worker.spec.ts
+++ b/src/app/modules/skstream/skstream.worker.spec.ts
@@ -1,5 +1,10 @@
-import { expect, describe, it, vi, afterEach } from 'vitest';
-import { apiGet, processVessel } from './skstream.worker';
+import { expect, describe, it, vi, afterEach, beforeEach } from 'vitest';
+import {
+  apiGet,
+  handleStreamEvent,
+  initVessels,
+  processVessel
+} from './skstream.worker';
 import { SKVessel } from '../skresources/resource-classes';
 
 // getVesselTrail() fetches the server-side "self" track with several apiGet()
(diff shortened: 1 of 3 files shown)


joelkoz:
@coderabbitai review

joelkoz:
@coderabbitai re the nitpick on `handleStreamEvent`'s parameter type — leaving it out of this PR deliberately.

That signature is pre-existing untyped code; this PR only adds `export` to it so the spec can drive it directly (the seam pattern already used for `apiGet` and `processVessel` in this worker). Typing it properly means modelling four heterogeneous shapes in one discriminated union: `onConnect` / `onClose` / `onError` carry DOM events — `onConnect` is read as `msg.target.readyState`, the other two ignore `msg` entirely — while `onMessage` carries a Signal K hello / response / delta payload that `parseStreamMessage` narrows itself downstream. Getting that union right would also force casts into the spec, where the close and error cases legitimately pass no message at all.

So it is a `refactor(stream):` change with its own review surface rather than part of a watchdog fix, and this repo asks for one logical change per PR. Happy to see it done separately.

Issue #686, opened 2026-08-08
Clear the 413 ESLint violations exposed by the flat-config migration, then gate CI on lint


joelkoz:
#684 restored a working ESLint setup (flat config + a `lint` script) but deliberately left the resulting violations alone, so that change touched no `src/` files. This issue tracks clearing them and finishing the job by gating CI.

`npm run lint` currently reports **413 problems (341 errors, 72 warnings) across 104 files** — 403 in `src/app`, 10 in `helper/`. Rule selection matches the pre-existing config (`eslint:recommended` + `@typescript-eslint` recommended, `eqeqeq` as a warning), so none of this comes from newly-introduced rules; it is simply what has accumulated while ESLint was silently broken.

### Breakdown

| Count | Rule |
|---:|---|
| 161 | `@typescript-eslint/no-explicit-any` |
| 94 | `@typescript-eslint/no-unused-vars` |
| 57 | `eqeqeq` *(warning)* |
| 39 | `prefer-const` |
| 17 | `no-useless-assignment` |
| 15 | unused `eslint-disable` directives *(warning)* |
| 11 | `no-case-declarations` |
| 4 | `@typescript-eslint/no-unused-expressions` |
| 4 | `no-constant-binary-expression` |
| 3 | `no-empty` |
| 2 | `preserve-caught-error` |
| 2 | `no-useless-catch` |
| 1 each | `no-unsafe-finally`, `no-constant-condition`, `no-unexpected-multiline`, `@typescript-eslint/no-namespace` |

36 errors and 22 warnings are auto-fixable with `--fix`.

### Suggested phasing

Landing this as one change would touch over a hundred files and mix mechanical edits with judgement calls, making it effectively unreviewable. Splitting by risk:

1. **Possible real defects — review individually.** `no-unsafe-finally`, `no-constant-binary-expression`, `no-constant-condition`, `no-unexpected-multiline`, `preserve-caught-error`, `no-useless-assignment` (~26 total). These rules exist because they catch genuine bugs; each wants a look rather than a blanket fix.
2. **Mechanical, low-risk.** `prefer-const`, `no-case-declarations`, unused `eslint-disable` directives, and the auto-fixable subset. Largely `--fix` territory, but the result still needs reading.
3. **`no-unused-vars` (94).** Mostly mechanical, though some will be unused parameters that want an `_` prefix convention rather than deletion.
[...]

PR #720, opened 2026-09-01
refactor(lint): clear no-unused-vars from #686

joelkoz:
Phase 3 of #686 — `@typescript-eslint/no-unused-vars`, 92 → 0. Total lint count
313 → 221.

Six commits, split by what kind of change each is rather than by file, since
"unused" covers everything from a dead import to a parameter an `effect()`
genuinely needs to keep reading.

### Imports and declarations (no runtime effect)

24 dead imports, plus two fully-unreferenced interfaces (`DrawEndEvent`,
`BuddyProperties`). Removing `DrawEndEvent` orphaned its only remaining
consumer's `Feature` import from `ol`, caught by the second lint pass and
removed alongside.

### Catch bindings and dead handler parameters

30 `catch (err)`/`catch (error)`/`catch (e)` bindings across 27 files, none
read in their block — replaced with the optional catch binding (`catch {}`,
ES2022, this project's target) rather than kept with a name nothing reads.
Applied by line, not by regex: several files mix live and dead bindings under
the same name, and a blind global replace broke the build in two files before
it was committed (`Cannot find name 'err'`), redone precisely against each
flagged line.

Two adjacent cleanups the same pass surfaced: `vessel-popover`'s two
`.subscribe(next, error)` callbacks never read their `HttpErrorResponse`
parameter; and `helper/alarms/alarms.ts`'s four route handlers declared
`next: NextFunction` and never called it — these are per-route handlers, not
`app.use` error middleware, so Express doesn't inspect their arity and
dropping the parameter changes nothing about how they're invoked.

### The judgment sites (11, each traced individually)

No blanket pattern applies here. Two worth flagging specifically:

- `alert.component`: `const ack = this.acknowledged()` inside an `effect()`
  looked unused but is **not** dead — the read is what registers the signal
  as an `effect()` dependency, independent of whether the value is stored.
  Deleting it would have silently narrowed what the effect reacts to (the
  exact bug class the effect/signal entry in `DEV-LESSONS-LEARNED.md` warns
  about). Kept as a bare read with a comment explaining why.
- `layer-s57-chart.component`: a computed `minZ`/`maxZ` looked like a missing
  `setMinZoom`/`setMaxZoom` call matching sibling raster/tilejson/vector-chart
  components. It isn't — the S-57 layer's actual zoom bounds are set
  independently inside `S57LayerStyler.ApplyStyle()`. Traced before deleting.

`fb-map.component`'s `onDragBoxCancel` and `linestyle-select`'s
`onSelectChange` (later commit) both dropped a now-unused event parameter that
their templates still passed `$event` to — Angular's strict template checking
treats that as a compile error, not a silent ignore, so the template call
sites are updated alongside.

### `argsIgnorePattern: '^_'` — reserved for override contracts, not general use

Some parameters exist only to satisfy an interface/override signature: a base
implementation's parameter is genuinely unused, but a sibling override reads
the same-named parameter (e.g. `sk-feature.component`'s
`onLabelZoomThreshold(entered)` is a no-op in the base; `layer-aisvessels`'
override reads `entered` to drive `toggleCogLines`). The parameter can't be
deleted there — the signature has to hold across the hierarchy — so it's
renamed to `_entered` and ESLint told to allow that pattern.

Scoped to `argsIgnorePattern` only, deliberately — a `_`-prefixed local
variable is still flagged. The convention is reserved for "this exists because
of a contract," not "I don't want to hear about this." Two base-class stubs
use it (`ais-base.component`, `sk-feature.component`); every other unused
parameter or binding in this PR is fully removed rather than renamed, since a
dropped parameter is unambiguous in a way a live, referenceable `_name`
binding isn't — nothing stops a future edit from actually calling it.

### One real defect found in passing, fixed separately

`resources.service.transformVessel()` has always taken an `id` parameter,
documented in its own JSDoc as "Vessel identifier," and returned an
`SKVessel` — whose base class declares `id: string` — without ever assigning
it. Confirmed dead since the file's earliest history, not a recent
regression. Both real call sites already have a meaningful id to hand it.
Traced every current consumer of the returned vessel to confirm nothing
depends on the field staying unset, so this closes a gap without being able
to break an existing read. Fixed rather than deleted, since the base class
already promises the field and both callers already have the value.

### One finding from the stand-in review, declined

CodeRabbit flagged the two `catch {}` blocks left empty in
`layer-bearing-line.component.ts` and `layer-cog-line.component.ts`
(`no-empty`). Both were already empty on `master` before this PR — this PR's
diff on those lines is purely the `catch (err) {}` → `catch {}` binding
removal `no-unused-vars` requires. `no-empty`'s total across the repo is
unchanged: 3 before, 3 after, same three sites. Declined as pre-existing and
outside this PR's scope; CodeRabbit agreed. `no-empty` (3, along with
`no-unused-expressions` 4 and `no-useless-catch` 2) never got its own phase in
#686's original breakdown and is worth a small follow-up on its own.

### Verification

630 tests / 70 files pass, `build:all` clean (webapp + helper), Prettier gate
clean on both `src/` and `helper/`.

### Noticed in passing, not addressed here

`settings.facade.buildFavouritesList()`: `const i = ...indexOf(...)` has been
computed and discarded since the file's oldest commit in this repo's history,
through three unrelated property renames. Deleting it changes nothing
observable today. Flagged as a possible abandoned feature rather than guessed
at — worth a look, not a lint fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vessel resource handling by consistently assigning vessel identifiers.
  * Preserved existing alarm, mapping, chart, weather, and resource behavior while simplifying event and error handling.

* **Refactor**
  * Removed unused parameters, imports, temporary values, and obsolete type references across the application.
  * Simplified settings, map interaction, alert, and resource-processing logic without changing visible functionality.

* **Chores**
  * Tightened linting rules to flag unused variables while allowing intentionally unused underscore-prefixed parameters.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -29,7 +29,11 @@ module.exports = [
       ...tsPlugin.configs['flat/recommended']
         .map((c) => c.rules)
         .reduce((all, rules) => Object.assign(all, rules), {}),
-      eqeqeq: 'warn'
+      eqeqeq: 'warn',
+      // A `_`-prefixed parameter is required by an override/interface
+      // contract that some other implementation genuinely reads, even
+      // though the base or this particular implementation does not.
+      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }]
     }
   }
 ];
--- a/helper/alarms/alarms.ts
+++ b/helper/alarms/alarms.ts
@@ -1,5 +1,5 @@
 // **** Signal K Standard ALARMS notifier ****
-import { NextFunction, Request, Response } from 'express';
+import { Request, Response } from 'express';
 import {
   ALARM_METHOD,
   ALARM_STATE,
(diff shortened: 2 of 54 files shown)


joelkoz:
Addressing the outside-diff finding on `autopilot.component.ts:368` (`autopilotApiPath`) in 69a89483.

Real miss, not something to argue with. `98c50193` removed `toggleEngaged()`'s only read of the field (the vestigial `uri` construction) but I didn't trace `autopilotApiPath`'s other references before committing — it was then write-only: declared, assigned in the constructor, never read. `no-unused-vars` doesn't catch unused private class fields the way it catches unused locals/params/imports (confirmed — ESLint reported nothing on this file both before and after), so nothing caught it until this review did.

Removed the field declaration and its constructor assignment. No remaining reference; no behaviour change beyond that.

joelkoz:
@coderabbitai review

PR #758, opened 2026-09-14
fix(charts): retry stalled or failed mapstyleJSON vector tiles

fakehec:
## What & why

A `mapstyleJSON` chart whose vector tiles are served over a high-latency or
flaky link (a boat's cellular/satellite uplink, in my case) can get **stuck
showing the previous zoom level and never recover on its own**.

The cause is in how OpenLayers handles a tile request that *stalls* — a
connection that is accepted but never delivers the body (as opposed to a clean
HTTP error):

- OpenLayers has **no application-level per-tile timeout**, so the tile sits in
  the `LOADING` state indefinitely.
- Because it believes the tile is still in flight, it **never re-requests it** —
  see [openlayers/openlayers#4338](https://github.com/openlayers/openlayers/issues/4338).
- Panning left/right or zooming out and back to the same level **reuses the
  stuck tile** instead of reloading it. Meanwhile the layer keeps rendering the
  over-zoomed *parent* tile from the level below, so the chart looks like it is
  "showing the previous zoom's projection" and refusing to refresh.

I reproduced this against a real online chart (Open Waters) in a headless
OpenLayers + ol-mapbox-style setup identical to Freeboard's: after stalling six
z-level tiles, zooming out and back and panning produced **zero** re-requests —
the map stayed on the coarse previous level. Adding the timeout+retry loader
below made the same scenario **self-heal with no user interaction**.

## The change

`makeChartTilesResilient(layerGroup)` installs a resilient tile load function on
every **vector tile source** the applied style creates (recursing into nested
groups). Each request gets a generous per-attempt timeout and is retried with
linear back-off; the tile is only marked `ERROR` once every attempt has failed.

Retrying happens **inside the loader, before the tile ever reaches the `ERROR`
state**, so it does not rely on a `tileloaderror` handler or on calling
`source.refresh()` — both documented to cause render loops with vector tile
sources ([openlayers/openlayers#17389](https://github.com/openlayers/openlayers/issues/17389)).
On a healthy connection the timeout never fires and behaviour is unchanged.

The component calls it from `MapStyleJsonChartLayerComponent` once `apply()`
resolves (when the sources exist).

## Scope

**Only vector tile sources are touched.** Raster (image) tile sources are left
on the OpenLayers default loader on purpose: the browser already surfaces image
load errors, and fetching image tiles through `fetch` rather than an ``
element would impose CORS requirements that many public raster tile servers do
not satisfy — a regression risk that isn't worth it for a case the browser
already handles.

Defaults are conservative: 30 s per-attempt timeout, 2 retries, 1 s base
back-off (all configurable via `ResilientTileLoadingOptions`).

## Testing

- `npm run build` — clean.
- `npm run lint` / `prettier --check` — clean.
- `npm test` — added 8 unit tests (retry logic under success / rejection /
  non-OK / stall-then-recover / exhaustion, and that `makeChartTilesResilient`
  hardens vector sources, skips raster sources, and recurses into nested
  groups). All pass; the existing `chart-utils` suite is untouched and still
  green.
- Functionally verified end-to-end in a headless OpenLayers harness against a
  live online vector chart (stall → stuck → self-heal after this change).

Independent of #733 (that PR normalises MapLibre styles for the OL renderer;
this one is about network resilience) — the two touch the same component but
address different problems.

https://claude.ai/code/session_01ET9sdhUU4uyew792nfV2Sa

## Summary by CodeRabbit

- **New Features**
  - Improved reliability for chart map tiles with automatic retries, timeouts, and backoff when loading fails.
  - Added resilient loading across nested vector tile layers.
  - Tile data is marked as failed only after all configured loading attempts are exhausted.

- **Bug Fixes**
  - Raster tile layers continue using their standard loading behavior.
  - Chart map layers can still be added when style or tile-resilience setup encounters an issue.
  - Improved handling when applying chart map styles through fallback loading.


joelkoz:
Thanks, Hugo — this is a good one. I checked the mechanism against the OL and olms sources and it holds up: OL's default vector loader (`ol/featureloader.js`) has no timeout and no retry, olms only installs its own tile load function when `transformRequest` is set (Freeboard doesn't use it, so nothing is lost by replacing the loader), and `apply()` resolves only after `finalizeLayer` has seen the source set, so hooking in after it is the right moment. It's also worse than you describe: `maxTilesLoading` in `ol/TileQueue.js` (default 16) is **map-wide** and a stalled tile holds its slot, so sixteen stalls on one online chart starve tile loading for every layer on the map. Worth fixing.

A few things before merge:

**1. Don't retry 4xx.** The loader currently retries any non-OK response. Open Waters returns 204 for empty tiles so it's fine there, but plenty of vector tile servers (tileserver-gl, some PMTiles/CDN setups) return **404 for empty tiles** — against one of those every empty ocean tile becomes three requests plus 3 s of back-off before it's marked `ERROR`, which triples traffic and adds latency on exactly the cellular link this is meant to help. A 4xx is a definitive answer: treat it like OL's default does today (straight to `ERROR`) and retry only network failures, aborts/timeouts, 5xx and 429. Please add a "404 is not retried" spec alongside the others.

**2. Sequencing with #733.** Both PRs edit the same `apply(...)` line in the component. #733 is further along, so let's land that first; then rebase this one and call `makeChartTilesResilient` from #733's `applyStyle` — on **both** paths (normalised style and the URL fallback), or the fallback silently loses resilience. While you're there, replace the bare `.catch(() => undefined)` with a `console.warn` so an `apply()` rejection isn't swallowed (that's the convention here for recoverable failures; `applyStyle` in #733 already does it).

**3. One behavioural test for the installed loader.** The `fetchArrayBufferWithRetry` specs are solid, but the `makeChartTilesResilient` ones only assert the load function's identity changed. CodeRabbit's nitpick is right: invoke the installed loader with a `VectorTile` stub and assert it reaches `TileState.ERROR` after exhaustion (and sets features on success).

CodeRabbit's `undefined`-overrides-defaults note is a trivial one — worth taking in the same push.

I'll approve the CI run now. Since this waits on #733, no rush — get that one over the line first and then rebase this on top.

fakehec:
Thanks for the review. Addressed in fa2ef3e9:

- **`fetchArrayBufferWithRetry` — explicit `undefined` overriding defaults:** now normalises each option with a nullish fallback (`options?.timeoutMs ?? DEFAULT…`, etc.), so `{ retries: undefined }` / `{ timeoutMs: undefined }` / `{ backoffMs: undefined }` keep the defaults instead of disabling retries, zeroing the timeout, or producing a NaN back-off.
- **Silent `.catch(() => undefined)` in the component:** the catch now reports the apply/harden failure via `console.warn` instead of swallowing it.

On the nitpick (behavioural test of the installed loader): I left the `makeChartTilesResilient` tests as wiring assertions (a `VectorTileSource` gets a new load function, raster sources don't, nested groups recurse). The fetch/parse/retry-then-`ERROR` behaviour is already covered by the `fetchArrayBufferWithRetry` unit tests, and exercising the *installed* loader end-to-end needs a real OpenLayers `VectorTile` load lifecycle (tile grid + projection + renderer) that isn't worth the flakiness in a unit test. Happy to add one if you'd prefer it.

Build, lint and Prettier are clean; the added `chart-utils` tests pass.

joelkoz:
Thanks for taking the CodeRabbit items, Hugo — and #733 is now merged, so this one can move.

Two things are still open from my review above (I think they may have been missed — the push only covered CR's comments):

**1. Don't retry 4xx.** `if (!response.ok) throw` still retries every non-OK response. Open Waters returns 204 for empty tiles so it's invisible there, but tileserver-gl and some PMTiles/CDN setups return **404 for empty tiles**, and against those every empty ocean tile becomes three requests plus 3 s of back-off before it's marked `ERROR` — more traffic and more latency on exactly the cellular link this is meant to help. A 4xx is a definitive answer: send it straight to `ERROR` like OL's default loader does today, and retry only network failures, aborts/timeouts, 5xx and 429. Please pin it with a "404 is not retried" spec next to the "503 then ok" one.

**2. Rebase onto master.** #733 changed the same `apply(...)` line into `applyStyle()`, so this branch now conflicts. After rebasing, call `makeChartTilesResilient` from `applyStyle` on **both** paths — the normalised-style `apply` and the URL fallback in the `catch` — otherwise the fallback silently loses resilience. (`applyStyle` already has the `console.warn`, so the one you added can fold into it.)

On the behavioural-loader test: fair enough, your rebuttal stands — the retry logic is covered and the loader is thin wiring. No need for that one.

Once that's pushed and CI is green this is good to go.

fakehec:
Rebased on master now that #733 is in, and addressed the review:

- **Wired into `applyStyle`, both paths.** `makeChartTilesResilient(layer)` is now called from #733's `applyStyle` after the normalised `apply()`, and also in the URL-fallback branch (which now awaits its `apply()` and hardens too, so the fallback doesn't silently lose resilience). The bare `.catch(() => undefined)` is gone — failures go through `console.warn`, matching the convention there.
- **4xx is not retried.** Only network errors, aborts/timeouts, 5xx and 429 retry; a 4xx (e.g. 404-for-empty-tile servers) fails immediately. Added a spec asserting 404 makes exactly one request while 503/429 retry.
- **Behavioural loader test.** Beyond the identity checks, two tests now invoke the installed loader against a `VectorTile` stub and assert it sets features on success and reaches `TileState.ERROR` after retries are exhausted.
- Kept CodeRabbit's per-field nullish defaults so an explicit `{ retries: undefined }` can't disable retries.

Also worth noting from testing against the boat: `maxTilesLoading` in `TileQueue` is map-wide (16), so a handful of stalls on one online chart really do starve every layer — folded that into the commit message.

Build, lint and Prettier clean; the added `chart-utils` tests pass. No rush on my end.

PR #762, opened 2026-09-16
fix(ais): honour the track minimum-zoom gate and parse tracks as line arrays

### LINKED ISSUE #706: AIS/aircraft track minimum-zoom gate has never applied
## Summary

The minimum-zoom gate for AIS and aircraft tracks has never applied. Two
independent defects in `layer-aistargets-track.component.ts` disable it, so
tracks render at any zoom regardless of `tracksMinZoom` (bound to `10` in
`fb-map.component.html`).

## The defects

**1. The guards test the method reference instead of calling it** — lines 95 and
102:

```ts
okToRenderTracks() {
  return this.mapZoom >= this.tracksMinZoom;
}
...
if (this.okToRenderTracks) {          // always truthy — a function reference
  this.onUpdateTargets(this.extractKeys(this.tracks));
}
```

A method reference is always truthy, so both guards pass unconditionally and
`addTrackWithId()` creates track features below the threshold.

**2. `buildStyle()` used to compute a zoom-based opacity, but the line was
dead.** Until #75 it read:

```ts
let color =
  this.mapZoom < this.tracksMinZoom ? `rgba(${rgb},0)` : `rgba(${rgb},1)`;
color = this.showTracks ? `rgba(${rgb},1)` : `rgba(${rgb},0)`;
```

The second statement overwrites the first unconditionally, so the zoom-derived
value never reached a caller. `no-useless-assignment` flagged it and #75 removed
it as dead code — behaviour-neutral, since it had never had any effect. Both
lines have coexisted since the component was first written.

`ngOnChanges` does still call `this.source.clear()` when the zoom crosses below
`tracksMinZoom`, so tracks briefly disappear — but with the guards above broken,
the next track update re-adds them.

## Why it was not fixed in #75

#75 is scoped to one ESLint rule group from #686 and is behaviour-neutral. This
is not an ESLint finding, and repairing it is a **user-visible change**: AIS and
aircraft tracks would begin disappearing below zoom 10, which no user has ever
seen because the gate has never worked. That deserves its own review rather than
riding along in a lint pass.

## Suggested fix

Call the method in both guards:

```diff
-    if (this.okToRenderTracks) {
+    if (this.okToRenderTracks()) {
```

Worth deciding at the same time whether `buildStyle()` should also factor the
zoom in (belt and braces), or whether the guards alone are the intended
mechanism.

Also worth confirming the zoom-10 default is still the behaviour we want, given
it has been inert for the component's entire history — turning it on is
effectively a new feature.

joelkoz:
## Summary

Two independent defects in `layer-aistargets-track.component.ts` kept AIS / aircraft tracks from rendering correctly. Both are in the same component and both are "tracks don't render as designed", so they ship together; a spec guards each one independently.

### 1. The minimum-zoom gate never applied (#706)

Both render guards tested the method *reference* — `if (this.okToRenderTracks)` — which is always truthy, so `tracksMinZoom` (bound to `10` in `fb-map.component.html`) had no effect and tracks were added at any zoom. The guards now call the method.

**User-visible change:** AIS and aircraft tracks now hide below zoom 10 and reappear at/above it, as the component was always meant to behave.

### 2. `parseCoordinates()` read the track one level too deep — tracks have thrown since v3.1.0

A track is a MultiLineString's coordinates (`SKVessel.track: Array<Position[]>` — `appendTrack()` pushes `[position]`, the tracks plugin returns `.coordinates`). The parser treated it as an array of MultiLineStrings and handed single `[lon, lat]` positions to `mapifyCoords()`. Before #570 that was a silent no-op (`Math.abs(undefined) >= 170` → false); since #570's `[...c]` spread it throws:

```
TypeError: c is not iterable
  at mapifyCoords (util.ts:65)
  at parseCoordinates (layer-aistargets-track.component.ts)
```

So on v3.1.0+ no AIS track could render at any zoom. It went unreported because `aisShowTrack` is off by default. The parser now iterates the lines once, matching `layer-tracks.component.ts`.

## Tests

New `layer-aistargets-track.component.spec.ts` (9 tests) on a bare prototype instance with a real OL `VectorSource`:

- `okToRenderTracks()`, `reloadTracks()` and `onUpdateTargets()` add no features below `tracksMinZoom` and do at/above it (incl. a non-default threshold)
- `parseCoordinates()` projects each line of an `Array<Position[]>` track, unwraps a dateline-crossing line, and produces a `MultiLineString` feature

Verified red→green: 8/9 fail on the unfixed component; with only the parse fix applied the 3 below-zoom tests still fail (`expected length 0, got 1`). Full suite 681/681, `build:web` and lint clean.

Closes #706

## Summary by CodeRabbit

- **Bug Fixes**
  - Track layers now respect the configured minimum zoom level when rendering and updating targets.
  - Track geometries crossing the international dateline are displayed correctly without spanning the entire map.
  - Multi-line target tracks are rendered with the correct geometry type.

- **Tests**
  - Added coverage for zoom-based rendering, target updates, coordinate projection, dateline handling, and multi-line track rendering.


joelkoz:
@coderabbitai review

